PR #34, opened 2015-09-12
new db method and improved wordage

chrispecoraro:
adding in a public method:
getAverageGlucoseReadingForLastMonth (returns Integer)
and improving the grammar in a text string.


bkerensa:
This looks good @chrispecoraro congrats on your first pull!

PR #76, opened 2015-10-01
correct and update ru translation strings

dmvstar:
correct and update ru translation strings for <array name="tips_all">


bkerensa:
@dmvstar This change should be made on http://translate.glucosio.org

dmvstar:
Ok
Thanks
When you planed merging translation to main branch (for example - ru translation is completed)

bkerensa:
@dmvstar Probably on Sunday as we are waiting for other languages to be merged in and we do it all at once with one push.

dmvstar:
Thanks, waiting.

Issue #134, opened 2015-10-26
Advance Graphing


bkerensa:
In order to show metrics from Issue #109 feature we will need to add those metrics in the graph or adjust how we do graphing so that it visualizes those other entry types to show how these events and other activities like carb intake or weight or medication or exercise impacts impacts glucose.

PR #137, opened 2015-10-28
Fix incorrect weekly average calculation for issue #134

lgp171188:
Always include and count the current week and its readings in the averages even
if it is incomplete. For example, if there are n readings in the current week which is incomplete, the weekly average should show the average including those current readings.


PaoloRotolo:
Looks good, thanks for the pull request.

PR #142, opened 2015-10-29
Reformat spaces and tabs

amouly:
Just a minor reformat on a helper class. We should follow the standard from Android Studio.


PaoloRotolo:
@amouly Merged, thanks!

PR #157, opened 2015-11-05
Swipe to remove functionality depricated at Assistant fragment

sch00lb0y:


PaoloRotolo:
Hi @sch00lb0y! I've tried your code: Assistant items doesn't swipe at all.
Can you explain why do you want to remove this functionality please?

PR #159, opened 2015-11-08
[Analytics] Relocated tracker ID

amouly:
## Changes

- Replace the Google Analytics Tracker ID from an String on the Application class to a BuildField placed on the Gradle configuration file. This allow us to have different tracker IDs for different types of builds. Right now apparently it's disabled for Debug mode, but we can use this technique for "release", "beta" or whatever flavor.
- Added some debug flags to Gradle file so it's easier to test from Studio.


PaoloRotolo:
LGTM, thanks.

Issue #52, opened 2015-09-14
Investigate adding Balance Rewards Support


bkerensa:
Once we start integrating third-party services and have support for recording weight, activity, bp etc we should investigate adding Balance Rewards Support.

https://developer.walgreens.com/api/balance-rewards/apimethod/balance-rewards-api

Since this wont be till way down the line I wont target this for now

PR #163, opened 2015-11-12
Created UI tests for Hello Activity

amouly:
# Description of Changes

- Refactor some IDs on HelloActivity layout, so now it's more descriptive and standard.
- Replace some "fill_parent" with "match_parent". From Android 2 this both attributes are the same, but the second one is the most used. Anyway, there is no visual changes.
- Added UI tests for HelloActivity
 - Check if every element is showing.
 - Check the submission flow.
- Created a custom ClickAction class due to some issues on Espresso library.
- I've found some issues when looking for the custom spinner on the UI hierarchy, so I've added an ID to it.

# Testing

These tests were run on:
> LG Nexus 5 - Android 5.1.1.

For testing it, you should connect a real device to your computer, and then run the Espresso tests with the following command:
> ./gradlew connectedAndroidTest

After that you can check the output from the console or the HTML file located in:
> {PROJECT_PATH} /app/build/reports/androidTests/connected/org.glucosio.android.html

### DIFF
--- a/app/src/main/res/layout/activity_hello.xml
+++ b/app/src/main/res/layout/activity_hello.xml
@@ -5,20 +5,21 @@
     android:id="@+id/helloactivity_mainframe"
     android:layout_width="match_parent"
     android:layout_height="match_parent"
-    android:visibility="visible">
+    android:visibility="visible"
+    tools:context="org.glucosio.android.activity.HelloActivity">
 
     <LinearLayout
         android:layout_width="match_parent"
-        android:layout_height="match_parent"
+        android:layout_height="wrap_content"
         android:orientation="vertical"
         android:paddingBottom="32dp"
         android:paddingLeft="32dp"
         android:paddingRight="32dp"
-        android:paddingTop="32dp"
-        tools:context="org.glucosio.android.activity.HelloActivity">
+        android:paddingTop="32dp">
 
         <TextView
-            android:layout_width="wrap_content"
+            android:id="@+id/activity_hello_title"
+            android:layout_width="match_parent"
             android:layout_height="wrap_content"
             android:paddingTop="32dp"
             android:text="@string/helloactivity_headline"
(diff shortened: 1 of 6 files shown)


app/src/main/res/layout/activity_hello.xml
```diff
@@ -5,20 +5,21 @@
...
     android:layout_width="match_parent"
     android:layout_height="match_parent"
-    android:visibility="visible">
+    android:visibility="visible"
+    tools:context="org.glucosio.android.activity.HelloActivity">
 
     <LinearLayout
         android:layout_width="match_parent"
-        android:layout_height="match_parent"
+        android:layout_height="wrap_content"
```
amouly: This change was proposed by Studio Lint. Should improve performance on the ScrollView.

PaoloRotolo:
Looks good, thanks!

amouly:
@piotrek1543 what do you think about this tests I've made? Do you want to review or improve it? Thank you.

piotrek1543:
@amouly Hi, your tests looks good. I added some notes about what I've already learned when I'd written tests for googlesamples repo called android-topeka.

1)Avoid copy-pasting test codes like you have if something on screen is displayed. Tests should check only one thing, so why don't you check if all parts are displayed using "for" loop.
2) If you have methods in one void with same id always connect them like you'd done in 15th test.
3) Tests should run independent, so adding prefix like check_007_ is not good practice.
4) Good practice would be using prefixes like <em>toolbar_checkIfTitleIsDisplayed</em> or <em>clickOnStartButton</em> 
5) In 15th you check two independent things ifChecked and ifUnchecked. Separate them for much clean code
6) I don't see it in your code, but never do actions checking transition from one activity to another. Like click on button Next, check if nextActivityToolbar is Displayed.

Check my closed PR on android-topeka repo to learn more: [#52](https://github.com/googlesamples/android-topeka/commit/f906b1ee19d75d015051a13e86f296eaf7496350)

PaoloRotolo:
Thanks for the review @piotrek1543. Feel free to improve it when you've time :)

PR #194, opened 2016-01-06
minify and shrinkResources in release builds

Aeyoun:


bkerensa:
This fails red

PaoloRotolo:
Minify isn't supported by some libraries we use. That's why it fails.

bkerensa:
Can we check with those libraries to see if they can add support? Does minify benefit us?

PaoloRotolo:
Realm has some problems with it. I'll try to search if there's a workaround.

Aeyoun:
I did actually write a long description when submitting this issue. Must have gotten lost in the void. The crux of it was that I tried getting this to work unsuccessfully but wanted to open a pull request as it was worth investigating. Hoped you guys who knew the code better would have better ideas on how to achieve this. Minify just gets the app to a place where zip alignment makes sense and would boost start-up performance and lower memory usage.

bkerensa:
Excellent we will look into this @PaoloRotolo can we file a bug upstream and see if they might make a fix?

PaoloRotolo:
Fixed in 0.9.2. You had to specify a custom ProGuard config file to find Realm proxy class after obfuscation.

```
-keep class io.realm.annotations.RealmModule
-keep @io.realm.annotations.RealmModule class *
-keep class io.realm.internal.Keep
-keep @io.realm.internal.Keep class * { *; }
-dontwarn javax.**
-dontwarn io.realm.**
```

PR #214, opened 2016-03-21
Introducing google flavour

emartynov:
I've created two flavour 'Google' and 'Amazon'. Most probably names should be changed to something more appropriate. I blindly moved all google play dependencies to it but would really like someone commented on it.

I also introduced robolectric tests and covered some of new functionality with it.

Please review and coment


emartynov:
Looks like I messed with git history. I used Android Studio "Rebase my GitHub fork" first time. I will try to clean things and create new pull request

Issue #244, opened 2016-06-29
App forces user to English language


emartynov:
Review from Google play:

> Where did the Czech? After updating disappeared Czech language.

emartynov:
We should add also language to `HelloActivity`. Otherwise, people who can not understand English will not pass it or will not be able to open Preferences

PaoloRotolo:
Yes after we fix the issue, the best option is delete the language settings in Database, then create a variable in SharedPreferences. Finally, check on start if the variable is created, so we don't delete data each time.

emartynov:
Will be released soon

PR #248, opened 2016-06-30
Fixing default locale bug #244

emartynov:
This is change to fix introduced issue #244.

The recipe is:
1. On the first start delete language from DB and remember it for future starts
2. Remove saving the wrong language from `HelloActivity`

Please review and comment

### DIFF
--- a/app/src/main/java/org/glucosio/android/GlucosioApplication.java
+++ b/app/src/main/java/org/glucosio/android/GlucosioApplication.java
@@ -92,9 +92,14 @@ public class GlucosioApplication extends Application {
         boolean cleanLocaleDone = preferences.isLocaleCleaned();
 
         if (!cleanLocaleDone) {
-            user.setPreferred_language(null);
+            User updatedUser = new User(user.getId(), user.getName(), user.getPreferred_language(),
+                    user.getCountry(), user.getAge(), user.getGender(), user.getD_type(),
+                    user.getPreferred_unit(), user.getPreferred_unit_a1c(),
+                    user.getPreferred_unit_weight(), user.getPreferred_range(),
+                    user.getCustom_range_min(), user.getCustom_range_max());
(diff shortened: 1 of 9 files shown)


PaoloRotolo:
It returns an exception at start:
```java
Caused by: java.lang.IllegalStateException: Changing Realm data can only be done from inside a transaction.
                                                       at io.realm.internal.Table.throwImmutable(Table.java:1402)
                                                       at io.realm.internal.Table.checkImmutable(Table.java:1074)
                                                       at io.realm.internal.UncheckedRow.setNull(UncheckedRow.java:281)
                                                       at io.realm.UserRealmProxy.realmSet$preferred_language(UserRealmProxy.java:154)
                                                       at org.glucosio.android.db.User.setPreferred_language(User.java:103)
                                                       at org.glucosio.android.GlucosioApplication.checkBadLocale(GlucosioApplication.java:95)
                                                       at org.glucosio.android.GlucosioApplication.initLanguage(GlucosioApplication.java:81)
                                                       at org.glucosio.android.GlucosioApplication.onCreate(GlucosioApplication.java:60)
```

The problem is the line ```user.setPreferred_language(null);``` that should happen inside a Realm transaction (read [here](https://realm.io/docs/java/0.80.0/#creating-objects)) because you're editing a Realm object.

To fix this, just create a new User object (not tied to Realm) and edit the values on it:
```
            User updatedUser = new User(user.getId(), user.getName(), user.getPreferred_language(),
                    user.getCountry(), user.getAge(), user.getGender(), user.getD_type(),
                    user.getPreferred_unit(), user.getPreferred_unit_a1c(),
                    user.getPreferred_unit_weight(), user.getPreferred_range(),
                    user.getCustom_range_min(), user.getCustom_range_max());
            updatedUser.setPreferred_language(null);
```

PaoloRotolo:
If you're ok we can merge now.

PR #262, opened 2016-07-10
Make Weight editable

HarKoss:
I fixed Weight edition and block edit button for other metrics as the functions are not yet implemented.
I started by weight, but I'll try to factorize for all of them.


PaoloRotolo:
Hey, I tried your fork but instead of **editing** a Weight reading, it just adds a new reading.
To reproduce:
 - Long click on a Weight item;
 - Select Edit;
- [ A new activity will appear but the Weight value is empty]
 - Add a new value and click save;
 - A new reading is created;

PaoloRotolo:
Also it would be cool to have edit available for other metrics as well (A1C, Ketones, Cholesterol...)

HarKoss:
@PaoloRotolo,
I saw the bug. If you edit the first weight entry (last in the display order), it's because this entry have an id=0 index in DB.
I thought about using date has index, like it's done for Glucose. But that will not solve the problem for already registered entry :/

satya164:
Don't we have `createTime` and `updateTime`  based on which we sort?

PaoloRotolo:
Hey @HarKoss! I've seen latest commit, it looks good for me. Can we update the other activities as well to extend ``AddReadingActivity`` and fix code duplication?

Some new tests were also written, @emartynov can you look at them please?

emartynov:
There are no new test :)

HarKoss:
Hi here,

@satya164, Yes it's sorted by Date when retrieving it from DB. But, it's referenced by id when you edit them.
I fixed that by adding a "editing" boolean.

@PaoloRotolo, It's just a beginning of refactoring. I'll try to make all reading editable in the row... :/

@emartynov, I even removed one ^^ it was just a try. I have to look at your test yes ! 
Peraps next commit will increase coverage. Who knows... ;-)

Thanks for your comment

PaoloRotolo:
So I'm merging this, since it fixes the issue for Weight. I'll open a new ticket for the other metrics.

Issue #250, opened 2016-07-02
Add language selector to Hello page


emartynov:
For devices or languages that don't pickup android locale. Becuase settings page is far away for this users

emartynov:
Pull request is merged. Closing issue

PR #266, opened 2016-07-23
Language selection on the Hello page

emartynov:
Changes for #250.

Added spinner for language selection on first application start.

The logic of filling spinner is same as logic is used on the Preference activity. New functionality covered with tests.

Please review.


PaoloRotolo:
Good job, nice you moved all the shared code in LocaleHelper.

PR #272, opened 2016-07-31
Fix default language selection on hello activity

emartynov:
Use locale to preselect language row on the device.

Please test, review and merge


PaoloRotolo:
The app is in italian when I open it, but the preselected language is still the first in the array 

screenshot_1470079290

emartynov:
@PaoloRotolo should work for you now

PaoloRotolo:
Looks good!

PR #278, opened 2016-08-08
Add Freestyle libre support

PaoloRotolo:


app/src/main/java/org/glucosio/android/activity/FreestyleLibre.java
```diff
@@ -0,0 +1,281 @@
...
+import org.glucosio.android.db.DatabaseHandler;
+import org.glucosio.android.object.PredictionData;
+import org.glucosio.android.object.ReadingData;
+import org.glucosio.android.tools.AlgorithmUtil;
+import org.glucosio.android.tools.AnimationTools;
+
+import java.io.IOException;
+import java.util.Arrays;
+
+public class FreestyleLibre extends Activity {
```
emartynov: Why not to have word Activity in the class name?

app/src/main/java/org/glucosio/android/activity/FreestyleLibre.java
```diff
@@ -0,0 +1,281 @@
...
+            // Start AddGlucose Activity passing the reading value
+            Intent intent = new Intent(getApplicationContext(), AddGlucoseActivity.class);
+            Bundle bundle = new Bundle();
+            String currentGlucose = mResult.history.get(mResult.history.size()-1).glucose(false);
+            bundle.putString("reading", currentGlucose + "");
+            intent.putExtras(bundle);
+            startActivity(intent);
+            FreestyleLibre.this.finish();
+        } else {
+            Intent intent = new Intent(getApplicationContext(), HelloActivity.class);
```
emartynov: So we will lose reading?

PaoloRotolo: @emartynov we can't add the reading because the user didn't create an user entry yet.
We can solve this passing the reading as extra to HelloActivity, then again to MainActivity and finally to AddGlucoseActivity. But as you can see this is a borderline case.

bkerensa:
Assigning to @emartynov for review

bkerensa:
Since we do not have a test device other than Paolo's sensor I am going to approve this into the wild and we can get some testing from the group of users who were already using the feature. 

I think it is pretty low risk that approving this will introduce a crasher or other user impacting bug.

PR #279, opened 2016-08-11
Update Build Gradle Depends

bkerensa:


PaoloRotolo:
Looks good, it's not building for a test failed (I think Eugen already fixed it in another branch).

PR #281, opened 2016-08-11
Update build.gradle depends

bkerensa:


emartynov:
I would not do dependencies update before code freeze. Who knows if it will introduce any issues. Ols dependencies were tested during entire cycle

emartynov:
Looks good to me!

PaoloRotolo:
I read Realm changelog and there're no breaking changes, only improvements. It's ok IMHO to merge.

PR #284, opened 2016-08-13
Update Google Play Strings File

bkerensa:
Would be good to have some eyes over this in case there are other typos, grammar or capitalization issues.


bkerensa:
Assigning @chwalters since he is a primary english speaker too

Issue #290, opened 2016-08-26
CSV export's layout is broken when exporting in mmol


PaoloRotolo:
From Glucosio's mail:
I have found the problem what i have while exporting CSV. When in app you set readings in mg/dl and export it to CSV, you get like you sent on the screenshot. 
But when readings in settings are set to mmol/ml then in export there is a problem like i have.

Thanx

PR #291, opened 2016-08-26
Add notes to CSV Export. Move save folder to Downloads. Perform additional checks before saving. Fix bug with mmol. Closes: #290.

PaoloRotolo:
Please review

### DIFF
--- a/app/src/main/java/org/glucosio/android/presenter/ExportPresenter.java
+++ b/app/src/main/java/org/glucosio/android/presenter/ExportPresenter.java
@@ -29,7 +29,6 @@ import android.os.Environment;
 import android.support.v4.app.ActivityCompat;
 import android.support.v4.content.FileProvider;
 import android.util.Log;
-import android.widget.Toast;
 
 import org.glucosio.android.activity.MainActivity;
 import org.glucosio.android.db.DatabaseHandler;
@@ -141,16 +140,6 @@ public class ExportPresenter {
     private boolean dirExists() {
         final File file = new File(Environment.getExternalStoragePublicDirectory(Environment.DIRECTORY_DOWNLOADS) + "/glucosio");
         return file.exists() || file.mkdirs();
-    }
-
-    private boolean arePermissionsGranted() {
-        if (hasStoragePermissions(activity)) {
-            File glucosioDirectory = new File(Environment.getExternalStoragePublicDirectory(Environment.DIRECTORY_DOWNLOADS) + "/glucosio");
-            glucosioDirectory.mkdirs();
-            return glucosioDirectory.exists();
-        } else {
-            return false;
-        }
     }
 
     private boolean hasStoragePermissions(Activity activity) {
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/glucosio/android/presenter/ExportPresenter.java
```diff
@@ -130,6 +138,21 @@ protected void onPostExecute(String filename) {
         }
     }
 
+    private boolean dirExists() {
+        final File file = new File(Environment.getExternalStoragePublicDirectory(Environment.DIRECTORY_DOWNLOADS) + "/glucosio");
+        return file.exists() || file.mkdirs();
+    }
+
+    private boolean arePermissionsGranted() {
```
emartynov: Not used method?

app/src/main/java/org/glucosio/android/tools/ReadingToCSV.java
```diff
@@ -42,72 +42,84 @@ public ReadingToCSV(Context mContext) {
...
 
+                // Concentration | Measured | Date | Time | Notes | Unit of Measurement
                 if ("mg/dL".equals(um)) {
                     for (int i = 0; i < readings.size(); i++) {
+                        GlucoseReading reading = readings.get(i);
+                        osw.append(dateTool.convertRawDate(reading.getCreated() + ""));
+                        osw.append(',');
 
-                        osw.append(readings.get(i).getReading() + "mg/dL");
+                        osw.append(dateTool.convertRawTime(reading.getCreated() + ""));
```
emartynov: Printing date time two times?

app/src/main/java/org/glucosio/android/tools/ReadingToCSV.java
```diff
@@ -42,72 +42,84 @@ public ReadingToCSV(Context mContext) {
...
                         osw.append(',');
 
-                        osw.append(readings.get(i).getReading_type() + "");
+                        osw.append(reading.getReading() + "");
                         osw.append(',');
 
-                        osw.append(dateTool.convertRawDate(readings.get(i).getCreated() + ""));
+                        osw.append("mg/dL");
```
emartynov: Are we really taking it in mg/dL? Or we actually store it in different formats?

emartynov:
Just minor comments. And I don't understand the fix again.

app/src/main/java/org/glucosio/android/tools/ReadingToCSV.java
PaoloRotolo: The first one converts the date. The second one the time:
```
osw.append(dateTool.convertRawDate(reading.getCreated() + ""));
osw.append(dateTool.convertRawTime(reading.getCreated() + ""));
```

PaoloRotolo:
So there were a typo in the CSV class when exporting with mmol as unit of measurement. The result was a broken CSV table. I fixed the typo and added new fields for Notes and other info. Also now the exported file goes in the Download folder (since on some devices we haven't permission to write outside).

app/src/main/java/org/glucosio/android/presenter/ExportPresenter.java
PaoloRotolo: Remove, thanks.

PR #292, opened 2016-08-27
Fixes Diabetes Type selector. Closes #288.

### LINKED ISSUE #288: Settings only able to display Type 1 and Type 2, Preferred range may also fail
When changing the diabetes type from Type 1 to either Gestational or LADA, Type 2 is always displayed as a result. Not sure how it't treated by the app, but it should of course display either Gestational or LADA.

This happens when e.g. type 2 was chosen during install. When choosing Gestational during install however, it shows "Type 3" afterwards in Settings, but then the same thing happens when changing it. Similar results occur when choosing LADA during install: Type 4 changes to Type 2 afterwards.

While at it, the Preferred range shows no type at all in settings, so can't see what's been selected, nor if that's remembered correctly, though it might fail since the Min and Max value don't change when selecting Custom range.

PaoloRotolo:
Please review.

### DIFF
--- a/app/src/main/java/org/glucosio/android/activity/PreferencesActivity.java
+++ b/app/src/main/java/org/glucosio/android/activity/PreferencesActivity.java
@@ -214,13 +214,23 @@ public class PreferencesActivity extends AppCompatActivity {
             diabetesTypePref.setOnPreferenceChangeListener(new Preference.OnPreferenceChangeListener() {
                 @Override
                 public boolean onPreferenceChange(Preference preference, Object newValue) {
-                    if (newValue.toString().equals(getResources().getString(R.string.helloactivity_spinner_diabetes_type_1))) {
+                    String[] typesArray = getResources().getStringArray(R.array.helloactivity_diabetes_type);
+                    String selectedType = newValue.toString();
+
+                    if (selectedType.equals(typesArray[0])) {
                         updatedUser.setD_type(1);
                         updateDB();
-                    } else {
+                    } else if (selectedType.equals(typesArray[1])) {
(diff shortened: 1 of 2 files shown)


emartynov:
I was reviewing but I didn't get point of the fix. What exactly is fixing the problem?

PaoloRotolo:
So originally there were only 2 diabetes types (1 and 2). Now we added other 2 types as well, but they were only added in the array in strings.xml. The selector in preferences had a if condition
```
if (selectedType.equals(typesArray[0])) {
   // set type 1
} else {
   // set type 2
}
```

I just added the other two types so they can be saved from preferences.

app/src/main/java/org/glucosio/android/activity/PreferencesActivity.java
```diff
@@ -214,13 +214,23 @@ public boolean onPreferenceChange(Preference preference, Object newValue) {
             diabetesTypePref.setOnPreferenceChangeListener(new Preference.OnPreferenceChangeListener() {
                 @Override
                 public boolean onPreferenceChange(Preference preference, Object newValue) {
-                    if (newValue.toString().equals(getResources().getString(R.string.helloactivity_spinner_diabetes_type_1))) {
+                    String[] typesArray = getResources().getStringArray(R.array.helloactivity_diabetes_type);
+                    String selectedType = newValue.toString();
+
+                    if (selectedType.equals(typesArray[0])) {
                         updatedUser.setD_type(1);
                         updateDB();
```
emartynov: You probably can call `updateDB()` after all if/else cases

emartynov:
LGTM

PR #295, opened 2016-09-10
Fix #285

### LINKED ISSUE #285: Export doesn't export notes
Export doesn't export notes

JoaquimLey:
- Fix Reading note exporting
- Abstract to a pseudo MVP architecture (ExportPresenter/ExportPresenterView)
- Improvements on export algorithm
- Simple unit tests
- Update gradle version

### DIFF
--- a/app/src/main/java/org/glucosio/android/presenter/ExportPresenter.java
+++ b/app/src/main/java/org/glucosio/android/presenter/ExportPresenter.java
@@ -74,7 +74,6 @@ public class ExportPresenter {
     public void onExportClicked(final boolean isExportAll) {
 
         if (hasStoragePermissions(mActivity)) {
-//            final ReadingToCSV csv = new ReadingToCSV(GlucosioApplication.getInstance());
             final String preferredUnit = dB.getUser(1).getPreferred_unit();
             final boolean[] isEmpty = {false};
 
--- a/app/src/main/java/org/glucosio/android/presenter/ExportPresenter.java
+++ b/app/src/main/java/org/glucosio/android/presenter/ExportPresenter.java
@@ -30,13 +30,13 @@ import android.support.v4.app.ActivityCompat;
 import android.support.v4.content.FileProvider;
 import android.util.Log;
 
-import org.glucosio.android.activity.MainActivity;
+import org.glucosio.android.GlucosioApplication;
 import org.glucosio.android.db.DatabaseHandler;
 import org.glucosio.android.db.GlucoseReading;
 import org.glucosio.android.tools.ReadingToCSV;
+import org.glucosio.android.view.ExportView;
 
 import java.io.File;
-import java.util.ArrayList;
 import java.util.Calendar;
 import java.util.List;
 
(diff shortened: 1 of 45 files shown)


app/src/main/java/org/glucosio/android/presenter/ExportPresenter.java
```diff
@@ -56,84 +56,78 @@
...
 
-    public void onExportClicked(final Boolean all) {
-        final List<GlucoseReading> readings;
+    private Activity mActivity;
+    private ExportView mExportView;
+    private DatabaseHandler dB;
 
-        if (all) {
-            readings = dB.getGlucoseReadings();
+    public ExportPresenter(Activity activity) {
```
emartynov: Why not to have parameter for constructor only `ExportView` type? Why do we need to cast it?

app/src/main/java/org/glucosio/android/presenter/ExportPresenter.java
```diff
@@ -56,84 +56,78 @@
...
-            fromDate.set(Calendar.DAY_OF_MONTH, fromDay);
-
-            Calendar toDate = Calendar.getInstance();
-            toDate.set(Calendar.YEAR, toYear);
-            toDate.set(Calendar.MONTH, toMonth);
-            toDate.set(Calendar.DAY_OF_MONTH, toDay);
-            readings = dB.getGlucoseReadings(fromDate.getTime(), toDate.getTime());
+            throw new RuntimeException("ExportPresenter Activity must implement ExportView interface");
         }
+        dB = GlucosioApplication.getInstance().getDBHandler();
```
emartynov: Why singleton preferred instead of parameter?

app/src/main/java/org/glucosio/android/presenter/ExportPresenter.java
```diff
@@ -56,84 +56,78 @@
...
+    }
+
+    public void onExportClicked(final boolean isExportAll) {
+
+        if (hasStoragePermissions(mActivity)) {
+//            final ReadingToCSV csv = new ReadingToCSV(GlucosioApplication.getInstance());
+            final String preferredUnit = dB.getUser(1).getPreferred_unit();
+            final boolean[] isEmpty = {false};
```
emartynov: Would it be better to create simple structure {String //filename, boolean //isEmpty} instead of having several states to update?

app/src/main/java/org/glucosio/android/tools/ReadingToCSV.java
```diff
@@ -30,22 +30,15 @@
...
 
-    private Context context;
-
-    public ReadingToCSV(Context mContext) {
-        this.context = mContext;
-    }
-
-
-    public String createCSVFile(Realm realm, final ArrayList<GlucoseReading> readings, String um) {
+    public static String createCSVFile(Context context, Realm realm, final List<GlucoseReading> readings, String um) {
```
emartynov: Why static method instead of class?

app/src/main/java/org/glucosio/android/tools/ReadingToCSV.java
```diff
@@ -84,13 +77,13 @@ public String createCSVFile(Realm realm, final ArrayList<GlucoseReading> reading
                         osw.append(dateTool.convertRawTime(reading.getCreated() + ""));
                         osw.append(',');
 
-                        osw.append(reading.getReading() + "");
+                        osw.append(String.valueOf(reading.getReading()));
```
emartynov: Nice!

app/src/main/java/org/glucosio/android/tools/ReadingToCSV.java
```diff
@@ -30,22 +30,15 @@
 import java.io.File;
 import java.io.FileOutputStream;
 import java.io.OutputStreamWriter;
-import java.util.ArrayList;
+import java.util.List;
```
emartynov: Nice!

app/src/test/java/org/glucosio/android/presenter/ExportPresenterTest.java
```diff
@@ -0,0 +1,52 @@
...
+
+    @Mock
+    private GlucoseReading glucoseReadingMock;
+
+    @Before
+    public void setUp() throws Exception {
+        initMocks(this);
+        when(dbHandlerMock.getGlucoseReading(anyLong())).thenReturn(glucoseReadingMock);
+        when(glucoseReadingMock.getNotes()).thenReturn(MOCK_NOTE_FOR_TEST);
+        when(glucoseReadingMock.getCreated()).thenReturn(new Date(System.currentTimeMillis()));
```
emartynov: `new Date()` is already have current milliseconds as time

app/src/test/java/org/glucosio/android/presenter/ExportPresenterTest.java
```diff
@@ -0,0 +1,52 @@
...
+        when(glucoseReadingMock.getReading()).thenReturn(TEST_READING_VALUE);
+    }
+
+
+    @Test
+    public void shouldReturnNoteFromReading_WhenReadingHasNote() throws Exception {
+        assertThat(glucoseReadingMock.getNotes()).matches(MOCK_NOTE_FOR_TEST);
```
emartynov: Testing just prepared mocks doesn't have sense

app/src/test/java/org/glucosio/android/presenter/ExportPresenterTest.java
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    @Test
+    public void shouldReturnNotReturnNoteFromReading_WhenReadingHasNoNote() throws Exception {
+        when(glucoseReadingMock.getNotes()).thenReturn("");
+        assertThat(glucoseReadingMock.getNotes()).doesNotMatch(MOCK_NOTE_FOR_TEST);
```
emartynov: Same comment about mocks

emartynov:
@JoaquimLey, thank you for the contribution!

I reviewed and added some comments. Can you reply?

Can you also elaborate how the algorithm was improved?
How notes exporting was fixed?

app/src/main/java/org/glucosio/android/presenter/ExportPresenter.java
JoaquimLey: Because you need the Activity since you are handling the permissions in the presenter (it wouldn't be really my choice but see line 140) which requires to have an activity, the cast is to ensure said activity is implementing the required View interface.

app/src/main/java/org/glucosio/android/presenter/ExportPresenter.java
JoaquimLey: This way you don't have to pass the databaseHandler as a parameter since you are always using the one residing in the application context anyway. Using a singleton you ensure you are accessing the same reference.

app/src/main/java/org/glucosio/android/tools/ReadingToCSV.java
JoaquimLey: Its not required to instantiate the class/object for the current usage, you only wish to "consume" code residing in it, which really makes it more of a **helper** class.
That being said I should've declared the class final, will do. 
This way you ensure fewer resources since the system will create the reference in compile time instead of runtime.

http://stackoverflow.com/a/218761/3374428
http://stackoverflow.com/a/5181618/3374428

app/src/test/java/org/glucosio/android/presenter/ExportPresenterTest.java
JoaquimLey: You're right.

app/src/test/java/org/glucosio/android/presenter/ExportPresenterTest.java
JoaquimLey: Just bootstrapped, not actual testing.

app/src/main/java/org/glucosio/android/presenter/ExportPresenter.java
JoaquimLey: So you can handle all three cases: When export process was successful, export process was successful but the database is empty or export process was unsuccessful.

JoaquimLey:
The current algorithm was querying the database on the UI thread, ensure there were `nrNotes > 0` then re-run the same query on off the way thread to return the value (inside the async task) which made no sense.
The export per-se wasn't using the same database reference and was probably the cause of the notes not exporting.
I'm pushing a new commit regarding your comments.

app/src/main/java/org/glucosio/android/presenter/ExportPresenter.java
emartynov: Clear! It was there already. As for me, a presenter should not have any dependency to Android API, but this is the discussion for the android team. 

As the quick solution I would have two parameters - one view and one activity. And remove casting!

app/src/main/java/org/glucosio/android/presenter/ExportPresenter.java
emartynov: This is actually breaking testability. If you have parameter then you can easy mock it in unit tests. You have to make static setters or some hooks with the singleton pattern.

app/src/main/java/org/glucosio/android/tools/ReadingToCSV.java
emartynov: Clear!

I would not bother about performance unless you show me it is the bottle neck or some problem here. Because this again breaks testability. With this solution, I can not mock this behaviour in unit tests. So if I test any class that uses this functionality, I will also test real exporting as well.

An article about "love" to statics http://www.endran.nl/blog/death-to-statics/

emartynov:
@JoaquimLey thank you for saving the UI thread! This is really important.

I'm not bug expert of the Realm, but looks like many references to Realm shouldn't be a problem (http://stackoverflow.com/questions/27777948/how-to-correctly-use-realm).

Can you reproduce the issue? Can you debug and confirm cause with notes exporting?

app/src/main/java/org/glucosio/android/presenter/ExportPresenter.java
JoaquimLey: Yes it would be easy to abstract the Android apis from the Presenter if there was no Permission handling there (and it shouldn't).
With the current setup you'll be instantiating something like this:
`ExportPresenter(this, this);` not pretty, but refactoring this wouldn't make the PR with to many differences.

app/src/main/java/org/glucosio/android/presenter/ExportPresenter.java
JoaquimLey: Once again I've gone along the lines of what was already present. I think the constructor should remain with the ExportView as the only param, and have a setter for the dbh. Once again I've gone along the lines of what was present. 

> But all scope logic should be done outside of consumer class

You're right.

app/src/main/java/org/glucosio/android/tools/ReadingToCSV.java
JoaquimLey: I always worry about performance, this can of course be tested if we change the `AsyncTask` to a Factory class/public method (as it should be). 
With the current setup, no you can't because you are receiving the clicked event and create the `AsyncTask` there, not a good pattern. And the fix _per se_ would get verbose, this would be my first contribution I don't think is a good atitude and change a lot, it may confuse/upset current contributors.

> Every kitten grows up to be a cat. They seem so harmless at first--small, quiet, lapping up their saucer of milk. But once their claws get long enough, they draw blood, sometimes from the hand that feeds them.

That being said, static elements are awesome when used right, they go both ways, if we ignore performance we'll have the same problems ;).

JoaquimLey:
In my case there were cases were the notes would export and others not, I couldn't find a pattern but I believe this issue was more related to the query of the note, not the saving or writing to the CSV file.
Also regarding realm, you can indeed use multiple references, but with the current implementation 
you are creating a new `DatabaseHandler` each time you access it:
```
@NonNull
    public DatabaseHandler getDBHandler() {
        return new DatabaseHandler(getApplicationContext());
    }
```
Once again I thought this was outside of this issue scope, that's why I assumed this could be the problem, having concurrent data access.

emartynov:
@PaoloRotolo, if you follow the conversation, what do you think?

JoaquimLey:
Hello again, so you guys don't what the contribution(s)?

emartynov:
Actually completely opposite - we love contributions!

PaoloRotolo:
Hey, sorry I totally missed this pr. The code looks good, can you please fix the conflict on ```build.gradle```. I think you're using a different version.

JoaquimLey:
@PaoloRotolo fixed (squash to prevent commit verbosity).

PaoloRotolo:
Merged, thanks.

bkerensa:
Great work thanks!

JoaquimLey:
Happy to help! ;)

Issue #294, opened 2016-09-08
Open source libraries blocker when offline


JoaquimLey:
### Steps to reproduce
1. Enter the app
2. Go to _About Glucosio_
3. Turn off your internet connection
4. Click on _Open source libraries_

### Expected behaviour
User should be able to see which Open source libraries are open

### Actual behaviour
You get a no connection on the webview, user is not able to "Pull to Refresh" even if he's online, must exit and re-enter the Activity/Fragment.

**What language/locale is your device set to use?:**
English

**What measurement system are you using:** 
Non-important

**Are you using backups:** yes/no
Non-important

### Device configuration
**Device:**
Non-important
**Operating system:**
Android
**Operating system version:**
6.0.1
**Downloaded App From:**
PlayStore

alexsimo:
There should be something like a "no connection" view or just cache the html.
Which would be more suitable?

satya164:
We should just cache the HTML I think

alexsimo:
Ok! I'll get working on this issue tonight

bkerensa:
@alexsimo Thanks!

alexsimo:
[Pull request] (https://github.com/Glucosio/glucosio-android/pull/302) made.

PR #302, opened 2016-09-23
Improvement for WebView activity

alexsimo:
Refactored the LicenseActivity and converted to a generic one that knows how to open some external links.

### DIFF
--- a/app/src/main/java/org/glucosio/android/activity/ExternalLinkActivity.java
+++ b/app/src/main/java/org/glucosio/android/activity/ExternalLinkActivity.java
@@ -20,7 +20,9 @@
 
 package org.glucosio.android.activity;
 
+import android.content.Context;
 import android.content.DialogInterface;
+import android.content.Intent;
 import android.os.Bundle;
 import android.support.v7.app.AlertDialog;
 import android.support.v7.app.AppCompatActivity;
@@ -37,9 +39,18 @@ import org.glucosio.android.tools.network.GlucosioExternalLinks;
 
 public class ExternalLinkActivity extends AppCompatActivity implements ExternalViewPresenter.View {
 
+    private static final String TITLE_KEY = "TITLE_KEY";
+    private static final String URL_KEY = "URL_KEY";
     private ExternalViewPresenter presenter;
     private WebView webView;
     private Map<String, Integer> toolbarTitle;
+
+    public static void launch(Context context, String title, String url) {
+        Intent intent = new Intent(context, ExternalLinkActivity.class);
+        intent.putExtra(TITLE_KEY, title);
+        intent.putExtra(URL_KEY, url);
+        context.startActivity(intent);
+    }
 
     @Override
     protected void onCreate(Bundle savedInstanceState) {
--- a/app/src/main/java/org/glucosio/android/activity/ExternalLinkActivity.java
+++ b/app/src/main/java/org/glucosio/android/activity/ExternalLinkActivity.java
@@ -117,7 +117,7 @@ public class ExternalLinkActivity extends AppCompatActivity implements ExternalV
         toolbarTitle = new HashMap<>();
         toolbarTitle.put(GlucosioExternalLinks.PRIVACY, R.string.preferences_privacy);
         toolbarTitle.put(GlucosioExternalLinks.LICENSES, R.string.preferences_licences_open);
-        toolbarTitle.put(GlucosioExternalLinks.TERMS, R.string.preferences_licences_open);
+        toolbarTitle.put(GlucosioExternalLinks.TERMS, R.string.preferences_terms);
     }
 
     private void setToolbarTitle(String string) {
--- a/app/src/main/java/org/glucosio/android/activity/ExternalLinkActivity.java
+++ b/app/src/main/java/org/glucosio/android/activity/ExternalLinkActivity.java
@@ -24,6 +24,7 @@ import android.content.Context;
 import android.content.DialogInterface;
 import android.content.Intent;
 import android.os.Bundle;
+import android.support.annotation.NonNull;
 import android.support.v7.app.AlertDialog;
 import android.support.v7.app.AppCompatActivity;
 import android.view.MenuItem;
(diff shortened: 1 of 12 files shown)


bkerensa:
Paolo, please review

app/src/main/java/org/glucosio/android/activity/ExternalLinkActivity.java
```diff
@@ -0,0 +1,134 @@
...
+            getSupportActionBar().setDisplayHomeAsUpEnabled(true);
+            getSupportActionBar().setDisplayShowHomeEnabled(true);
+        }
+    }
+
+    private void initTitles() {
+        toolbarTitle = new HashMap<>();
+        toolbarTitle.put(GlucosioExternalLinks.PRIVACY, R.string.preferences_privacy);
+        toolbarTitle.put(GlucosioExternalLinks.LICENSES, R.string.preferences_licences_open);
+        toolbarTitle.put(GlucosioExternalLinks.TERMS, R.string.preferences_licences_open);
```
emartynov: Should be terms link I guess

app/src/main/java/org/glucosio/android/presenter/ExternalViewPresenter.java
```diff
@@ -0,0 +1,77 @@
...
+  private Map<String, String> urlMatchings;
+
+  public ExternalViewPresenter(View view, NetworkConnectivity network) {
+    this.view = view;
+    this.network = network;
+    initUrls();
+  }
+
+  private void initUrls() {
+    urlMatchings = new HashMap<>();
```
emartynov: You can init it with size 3

app/src/main/java/org/glucosio/android/presenter/ExternalViewPresenter.java
```diff
@@ -0,0 +1,77 @@
...
+      view.setupToolbarTitle(url);
+      view.loadExternalUrl(url);
+    } else {
+      view.showNoConnectionWarning();
+    }
+  }
+
+  private String matchUrlToAction(String action) {
+    String url = GlucosioExternalLinks.TERMS;
+    if (action != null && urlMatchings.containsKey(action)) {
```
emartynov: I'm a bit puzzled. Why we have logic of matching here and just don't pass url and title outside of this activity/presenter?

emartynov:
This also marked as fix for #294. As for offline, the changes will show a dialogue and will quit the activity (BTW I missed, `AlertDialog` is cancelable so if the user just presses back or tap out on the activity he will stay on it).

I was thinking that fix would be to include HTML files to the app. So no matter connectivity we will show them.

app/src/main/java/org/glucosio/android/activity/ExternalLinkActivity.java
alexsimo: Nop, the activity is used to load any kind of external links. I didn't change behavior, just refactored.

app/src/main/java/org/glucosio/android/presenter/ExternalViewPresenter.java
alexsimo: I don't use to init collections with their expected size, does it report any significant advantage?

app/src/main/java/org/glucosio/android/presenter/ExternalViewPresenter.java
alexsimo: I don't understand. All the logic should be in the presenter, the view doesn't have to know anything about logic and how it works, it's a _passive view_.

alexsimo:
> I was thinking that fix would be to include HTML files to the app. So no matter connectivity we will show them.

@emartynov yep, i was thinking on caching the html but in this case is loading all the website, not a simple plain html. Another option is having the html locally but normally the terms, licenses and privacy are stuff that can change a lot, if that happens, making a new release only changing some html would be silly.

Another aspect is that we are talking about terms screen. We shouldn't invest a lot in this screen, there are lot more apps parts that need love.

app/src/main/java/org/glucosio/android/activity/ExternalLinkActivity.java
emartynov: I mean that string should be probably `preferences_terms`

app/src/main/java/org/glucosio/android/presenter/ExternalViewPresenter.java
emartynov: Not in this case, but it is just my natural habit (http://stackoverflow.com/questions/2247773/is-it-worthwhile-to-initialize-the-collection-size-of-a-listt-if-its-size-rea)

emartynov:
@alexsimo, from my experience things like "terms & conditions" with "privacy notes" don't change often. You make them right first and change later will require a lot of dance - notify the user by official channel of communication, modifying the app to show them first and accepting them before any app usage.

app/src/main/java/org/glucosio/android/activity/ExternalLinkActivity.java
alexsimo: You're right. Duplicating lines..

app/src/main/java/org/glucosio/android/presenter/ExternalViewPresenter.java
alexsimo: Understood. In my daily work i don't use to predefine the expected collection size but nice! :)

app/src/main/java/org/glucosio/android/presenter/ExternalViewPresenter.java
alexsimo: > If we will have one more link to open then you have to modify two places - this presenter and place which calls ExternalLinkActivity.

I don't want to pollute presenter with implementation details like string ids.

> I would also instead of the dance with null add precondition that activity will crash if it is null.

Could be an option for a future PR. Actually i don't want to change any existing behaviour and limit the PR to the scope of the issue.

alexsimo:
Same as above comment. I would leave the offline saving for the terms for some future PRs.
As i don't want to change existing behaviour. And this kind of stuff should be asked by @bkerensa or some owner IMO.

satya164:
Could you enable app cache for the webview? We only have to add app cache on server for the app to able to cache it

alexsimo:
@satya164 i can, but after when into the activity i check the connection, it doesn't matter if there is any cache. I show a dialog and finish the activity.

Makes any sense having the webview cache with the actual behaviour?

satya164:
> @satya164 i can, but after when into the activity i check the connection, it doesn't matter if there is any cache. I show a dialog and finish the activity.

If there's appcache present, the webview will show the cached version. I'm not sure there is a way to check if cache is present or not, but I think it's fine since it only needs internet for the first connection. It'll show the cached version from there and update the cache when the manifest on server changes automatically (if internet is present).

To determine if it cached everything, we could use JS in the webview to notify when caching is complete, but I don't think it's worth it.

alexsimo:
> I'm not sure there is a way to check if cache is present or not

Me either :-/

> To determine if it cached everything, we could use JS in the webview to notify when caching is complete, but I don't think it's worth it.

We should not complicate something like terms and conditions. There are more important issues to be solved i think 😃 

@satya164, @emartynov and @PaoloRotolo  any thumbs for merging this

satya164:
> We should not complicate something like terms and conditions. There are more important issues to be solved i think

That's why I said I don't think it's worth it. We can just enable appcache and I think it's good enough

alexsimo:
@satya164 if there is no connection, the activity with the webview will never be showed. Would you still add the webview cache?

satya164:
@alexsimo If we enable cache, we won't need that check at all

alexsimo:
@satya164 true, but personally i would not add code that *by the moment* it won't be accessed
Let's just add the cache when the server will be prepared :)

app/src/main/java/org/glucosio/android/presenter/ExternalViewPresenter.java
emartynov: I don't ask you to pollute presenter.

I'm saying that design right now is next:
`some place launch web activity with state name -> web activity/web presenter trying from state name get url and title -> web page is show with title`

But more clean would be:
`some place launch web activity with url and title -> web activity shows url and title`

If you don't want to pollute code with urls and titles then you can write some helper class:
`WebActivityHelper {
  void launchTermsAndConditions(Context context);
 ...
}`
And this helper class will keep pairs of titles and urls.

alexsimo: Yep! Nice way! This way the activity / presenter will be as dumb as possible.
Tomorrow i will push more code. 

Thanks!!

bkerensa:
FWIW our wordpress web server is a managed service so we can't install any special caching packages. It's alresdy had a blend of caching technology and load balancing.

satya164:
@bkerensa we'll just need to add a static file. we can do that right?

bkerensa:
Maybe I'll have to check I think they only allow core wp files and themes and their back end scans constantly for risks.

We should consider just copying TOS to the app and not translating it.

emartynov:
Guys, let just include html/text files in the app

bkerensa:
Can we just to a string XML file? Is there any pro or con to using HTML vs XML file?

alexsimo:
@bkerensa with HTML we have easier formatting. With XML can also have but more work to do.

Attach the TOS here if you want and I'll modify the PR to the new requirements.

alexsimo:
Changed behaviour as @emartynov pointed me :)

app/src/main/java/org/glucosio/android/activity/ExternalLinkActivity.java
```diff
@@ -37,10 +39,19 @@
 
 public class ExternalLinkActivity extends AppCompatActivity implements ExternalViewPresenter.View {
 
+    private static final String TITLE_KEY = "TITLE_KEY";
+    private static final String URL_KEY = "URL_KEY";
     private ExternalViewPresenter presenter;
     private WebView webView;
     private Map<String, Integer> toolbarTitle;
 
+    public static void launch(Context context, String title, String url) {
```
emartynov: Add `@NonNull` annotations and preconditions

app/src/main/java/org/glucosio/android/activity/ExternalLinkActivity.java
```diff
@@ -89,13 +89,20 @@ public void setupToolbarTitle(String link) {
...
         if (getIntent().getExtras() != null) {
-            p = getIntent().getExtras();
-            return p.getString("key");
+            title = getIntent().getStringExtra(TITLE_KEY);
         }
-        return null;
+        return title;
+    }
+
+    @Override public String extractUrl() {
```
emartynov: That is small duplication. If you add extra parameter key then you can replace two methods with one

app/src/main/java/org/glucosio/android/presenter/ExternalViewPresenter.java
```diff
@@ -36,13 +36,24 @@ public void onViewCreated() {
...
   }
 
+  private void parametersPrecondition(String title, String url) {
+    if ((invalidParam(title)) || invalidParam(url)) {
+      throw new RuntimeException("Invalid arguments: need URL and TITLE");
+    }
+  }
+
+  private boolean invalidParam(String url) {
+    return url == null || url.length() == 0;
```
emartynov: https://developer.android.com/reference/android/text/TextUtils.html#isEmpty(java.lang.CharSequence)

app/src/main/java/org/glucosio/android/presenter/ExternalViewPresenter.java
```diff
@@ -36,13 +36,24 @@ public void onViewCreated() {
...
       view.setupToolbarTitle(title);
       view.loadExternalUrl(url);
     } else {
       view.showNoConnectionWarning();
     }
   }
 
+  private void parametersPrecondition(String title, String url) {
+    if ((invalidParam(title)) || invalidParam(url)) {
+      throw new RuntimeException("Invalid arguments: need URL and TITLE");
```
emartynov: I would throw `InvalidArgumetException`

app/src/main/java/org/glucosio/android/activity/ExternalLinkActivity.java
alexsimo: I don't get this one, sorry.
Can please give more explanation please?

emartynov: You have two methods `extractUrl` and `extractTitle` which can be replaced with one `getIntentExtraSafe(@NonNull String extraKey)`

alexsimo: But the presenter will need to know the _extraKey_ that should be public and static on the `ExternalLinkActivity` right?

emartynov: Agree, I didn't notice it is part of contract between presenter and view. So in this case it is matter of taste:
 1. Keep contract and have one general private method in activity wich knows about paramters
 2. Make parameters defined in contract and use them everywhere

As for me the first one is cleaner and preferrable

PaoloRotolo:
Code looks good. Just one thing, since we don't change licences and privacy info so often I don't think we need to spend a lot of time here. We just copy the plain text from the website and put it in a simple view, without WebViews. Then we make some changes we update it each release.

What do you think?

app/src/main/java/org/glucosio/android/presenter/ExternalViewPresenter.java
alexsimo: Done @emartynov. Thanks for the tips!

alexsimo:
@PaoloRotolo 😫  i would make another pr for that..

app/src/test/java/org/glucosio/android/presenter/ExternalViewPresenterTest.java
```diff
@@ -21,16 +22,19 @@
...
+  public void shouldThrowException_WhenNoParameters() throws Exception {
     when(network.isConnected()).thenReturn(true);
     when(view.extractTitle()).thenReturn(null);
 
-    presenter.onViewCreated();
-
+    try {
+      presenter.onViewCreated();
+    } catch (Exception ex) {
+      assertTrue(ex instanceof IllegalArgumentException);
```
emartynov: Why not `expected` in test annotations?

app/src/main/java/org/glucosio/android/activity/ExternalLinkActivity.java
```diff
@@ -39,115 +40,109 @@
...
-        if (getSupportActionBar() != null) {
-            getSupportActionBar().setDisplayHomeAsUpEnabled(true);
-            getSupportActionBar().setDisplayShowHomeEnabled(true);
-        }
+  private void setupToolbar() {
+    if (getSupportActionBar() != null) {
+      getSupportActionBar().setDisplayHomeAsUpEnabled(true);
+      getSupportActionBar().setDisplayShowHomeEnabled(true);
     }
-
```
emartynov: That is want I would name formatting issue. Why your code was not formatted before?

app/src/test/java/org/glucosio/android/presenter/ExternalViewPresenterTest.java
alexsimo: Due the codacy warning..

app/src/main/java/org/glucosio/android/activity/ExternalLinkActivity.java
alexsimo: I'm using the Square checkstyle and if I apply it it will differ from other files in the project.
This one I may apply it unintentionally.

emartynov: Clear! It happens. Here checkstyle will help immediately :)

app/src/test/java/org/glucosio/android/presenter/ExternalViewPresenterTest.java
emartynov: I'm thinking to switch it off. It gives too many false positives.

emartynov:
Nice changes for WebView Activity.

Next PR I hope we will include offline files and fix original issue.

Issue #300, opened 2016-09-23
Fixed 'Get started' button on user onboarding


alexsimo:
### Steps to reproduce
1. Open the application for first time
2. Onboarding / user registration screen is showed
3. You have to scroll to see the **Get started** button

### Expected behaviour
In terms of UX the button should be fixed in some kind of fixed static bar.
You shouldn't have to scroll to see the **Get started** button. In small screen devices can require some effort to figure out.

### Actual behaviour
You have to scroll to find the button.

PR #305, opened 2016-09-28
Pin get started button to the bottom of the on-boarding screen

deva666:
As requested in #300 , maybe this is how it can be solved. Get started button is pinned to the bottom of the screen and there's a border above it.
When an user scrolls all the way to the bottom, border disappears and then it looks like before.

### DIFF
--- a/app/src/main/java/org/glucosio/android/activity/HelloActivity.java
+++ b/app/src/main/java/org/glucosio/android/activity/HelloActivity.java
@@ -28,6 +28,7 @@ import android.support.v4.content.res.ResourcesCompat;
 import android.support.v7.app.AppCompatActivity;
 import android.util.Log;
 import android.view.View;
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/glucosio/android/activity/HelloActivity.java
```diff
@@ -141,6 +154,27 @@ private void initStartButton() {
         }
     }
 
+    private void initScrollViewListener() {
```
alexsimo: This could not be better achieved using a CoordinatorLayout?

bkerensa:
@elioqoshi hasn't yet started on our mocks so onboarding is still on hold

emartynov:
Marco, thank you for the contribution! I'm going to close PR since there is no activity in it. Sorry that it is because of delays on our side.

PR #306, opened 2016-09-28
Improve FreeStyle Libre support

kaindl:


app/src/main/java/org/glucosio/android/activity/FreestyleLibreActivity.java
```diff
@@ -176,7 +176,7 @@ private void showReadingLayout() {
 
         // Apply values in TextViews
         // TODO: Add check for mmol/L
-        readingTextView.setText(mResult.history.get(mResult.history.size()-1).glucose(false));
+        readingTextView.setText(mResult.trend.get(0).glucose(false));
 
         new Runnable() {
```
alexsimo: What's the reason of the runnable?

kaindl: didn't touch that yet and didn't check what is really needed, but there is imho a lot of code in the FSL-parts, we don't need/use. If it works good for a few people like this, i would start to clean up the code..

PaoloRotolo:
I don't have a FreeStyle to test it out, but I'm merging anyway.
Everything is better than the current implementation :D

PaoloRotolo:
Ah @kaindl can you fix the conflict please?

Issue #187, opened 2015-12-11
Allow comma as decimal separator


Aeyoun:
My device’s default keyboard, the Sony Xperia Keyboard, only has a comma but not period for use as a decimal separator. I’m in a region that uses commas instead of periods and naturally I’d appreciate to enter values with a comma. (The split between the two decimal separators is roughly [50%/50%](https://www.slightfuture.com/webdev/html5-input-number-localization.html) – the split goes between former British colonies plus China, and then everyone else).

The international standard is “1 000,00” (ISO 31-0) and the way discouraged by all standards bodies but is still used in half the world is variants on “1,000.00”. As all glucose levels any user would ever record values  less than a thousand, the thousand separator problem can be avoided and Glucosio can **accept periods and commas interchangeably in user input.** Assuming the right-most comma or period as the decimal separator and stripping any other periods and commas should also work.

TimothyCuffe:
Maybe add an option in the settings menu for Decimal separators? That way everything could easily be switched back and forth?

Aeyoun:
Should at least respect the locale setting. Though, ideally it should just accept both variants no questions asked. Would anyone really insert a thousand separator in the value ranges expected by this app?

betsythefc:
The only problem I see with this is the CSV export. We would need to surround all decimal numbers with quotes so that any spreadsheet/CSV reader will know it's not a new cell.

Aeyoun:
@betsythefc, this issue doesn’t affect how data is stored internally or exported. This is about the localization of the user interface and user provided input.

bkerensa:
@PaoloRotolo Lets try to land a fix for this and the CSV export decimal issue in our next release.

emartynov:
Addressed in #307. Fixed!

PR #307, opened 2016-10-01
Allow comma as decimal separator when adding Glucose reading

prt2121:
addresses https://github.com/Glucosio/glucosio-android/issues/187 (hacktoberfest)

note:
I use `NumberFormat#parse` to parse the reading value. 

if you prefer
> "Assuming the right-most comma or period as the decimal separator and stripping any other periods and commas should also work."

I will update my PR

thanks!

### DIFF
--- a/app/src/main/java/org/glucosio/android/presenter/AddGlucosePresenter.java
+++ b/app/src/main/java/org/glucosio/android/presenter/AddGlucosePresenter.java
@@ -81,25 +81,7 @@ public class AddGlucosePresenter extends AddReadingPresenter {
             if (number == null) {
                 activity.showErrorMessage();
             } else {
-                boolean isReadingAdded;
-                if ("mg/dL".equals(getUnitMeasuerement())) {
-                    int finalReading = number.intValue();
-                    GlucoseReading gReading = new GlucoseReading(finalReading, type, finalDateTime, notes);
-                    if (oldId == UNKNOWN_ID) {
-                        isReadingAdded = dB.addGlucoseReading(gReading);
-                    } else {
-                        isReadingAdded = dB.editGlucoseReading(oldId, gReading);
-                    }
-                } else {
-                    converter = new GlucosioConverter();
-                    int convertedReading = converter.glucoseToMgDl(number.doubleValue());
-                    GlucoseReading gReading = new GlucoseReading(convertedReading, type, finalDateTime, notes);
-                    if (oldId == UNKNOWN_ID) {
-                        isReadingAdded = dB.addGlucoseReading(gReading);
-                    } else {
-                        isReadingAdded = dB.editGlucoseReading(oldId, gReading);
-                    }
-                }
+                boolean isReadingAdded = createReading(type, notes, oldId, finalDateTime, number);
                 if (!isReadingAdded) {
                     activity.showDuplicateErrorMessage();
                 } else {
@@ -109,6 +91,24 @@ public class AddGlucosePresenter extends AddReadingPresenter {
         } else {
             activity.showErrorMessage();
         }
+    }
+
+    private boolean createReading(String type, String notes, long oldId, Date finalDateTime, Number number) {
+        boolean isReadingAdded;
+        int readingValue;
+        if ("mg/dL".equals(getUnitMeasuerement())) {
+            readingValue = number.intValue();
+        } else {
+            converter = new GlucosioConverter();
+            readingValue = converter.glucoseToMgDl(number.doubleValue());
+        }
+        GlucoseReading gReading = new GlucoseReading(readingValue, type, finalDateTime, notes);
+        if (oldId == UNKNOWN_ID) {
+            isReadingAdded = dB.addGlucoseReading(gReading);
+        } else {
+            isReadingAdded = dB.editGlucoseReading(oldId, gReading);
+        }
+        return isReadingAdded;
     }
 
     public Integer retrieveSpinnerID(String measuredTypeText, List<String> measuredTypelist) {
--- a/app/src/main/java/org/glucosio/android/presenter/AddGlucosePresenter.java
+++ b/app/src/main/java/org/glucosio/android/presenter/AddGlucosePresenter.java
@@ -39,6 +39,7 @@ import java.util.Date;
 import java.util.List;
 
 public class AddGlucosePresenter extends AddReadingPresenter {
+    private static final int UNKNOWN_ID = -1;
     private DatabaseHandler dB;
     private AddGlucoseActivity activity;
     private ReadingTools rTools;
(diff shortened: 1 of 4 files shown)


emartynov:
Thank you a lot for the contribution! Could you also write some tests that demonstrate user issue and it will be also good check for fix regression.

app/src/main/java/org/glucosio/android/presenter/AddGlucosePresenter.java
```diff
@@ -119,6 +132,20 @@ public void dialogOnAddButtonPressed(String time, String date, String reading, S
...
+    @Nullable
+    private Number parseReading(@Nullable String reading) {
+        if (reading == null)
+            return null;
+        NumberFormat numberFormat = NumberFormat.getInstance();
+        Number number = null;
+        try {
+            number = numberFormat.parse(reading);
+        } catch (ParseException e) {
+            activity.showErrorMessage();
```
emartynov: You will show in this case error message two times - one here and another immediate after when null arrives to the calling method.

app/src/main/java/org/glucosio/android/presenter/AddGlucosePresenter.java
```diff
@@ -72,14 +75,19 @@ public int hourToSpinnerType(int hour) {
     public void dialogOnAddButtonPressed(String time, String date, String reading, String type, String notes) {
         if (validateDate(date) && validateTime(time) && validateGlucose(reading) && validateType(type)) {
             Date finalDateTime = getReadingTime();
+            Number number = parseReading(reading);
+            if (number == null) {
+                activity.showErrorMessage();
+                return;
+            }
             boolean isReadingAdded;
```
prt2121: yes sure. will do

app/src/main/java/org/glucosio/android/presenter/AddGlucosePresenter.java
```diff
@@ -119,6 +132,20 @@ public void dialogOnAddButtonPressed(String time, String date, String reading, S
         }
     }
 
+    @Nullable
+    private Number parseReading(@Nullable String reading) {
```
prt2121: it shouldn't be `null` from `dialogOnAddButtonPressed` but i put it there in case it'll be used in other places in the future.

app/src/main/java/org/glucosio/android/presenter/AddGlucosePresenter.java
prt2121: ah. my bad. that got in when i refactored my code.

prt2121:
thanks for your review. i'll fix this soon

app/src/main/java/org/glucosio/android/presenter/AddGlucosePresenter.java
emartynov: YAGNI?

prt2121: yes, sure. i'll remove it. thx!

app/src/main/res/layout/activity_add_glucose.xml
```diff
@@ -90,6 +90,7 @@
                             android:layout_width="fill_parent"
                             android:layout_height="wrap_content"
                             android:hint="@string/dialog_add_concentration"
+                            android:digits="0123456789.,"
```
emartynov: I don't understand, this bit of change. Is it required?

app/src/main/java/org/glucosio/android/presenter/AddGlucosePresenter.java
```diff
@@ -70,49 +71,40 @@ public int hourToSpinnerType(int hour) {
...
+                    GlucoseReading gReading = new GlucoseReading(finalReading, type, finalDateTime, notes);
+                    if (oldId == UNKNOWN_ID) {
+                        isReadingAdded = dB.addGlucoseReading(gReading);
+                    } else {
+                        isReadingAdded = dB.editGlucoseReading(oldId, gReading);
+                    }
+                } else {
+                    converter = new GlucosioConverter();
+                    int convertedReading = converter.glucoseToMgDl(number.doubleValue());
+                    GlucoseReading gReading = new GlucoseReading(convertedReading, type, finalDateTime, notes);
```
emartynov: Line 96 till 101 is duplication for line 87 till 92. You can avoid it if you move them out of `if/else` and just keep in else preparing int value for `Reading`,

app/src/main/java/org/glucosio/android/presenter/AddGlucosePresenter.java
```diff
@@ -70,49 +71,40 @@ public int hourToSpinnerType(int hour) {
...
-            }
-            if (!isReadingAdded) {
-                activity.showDuplicateErrorMessage();
+            Number number = ReadingTools.parseReading(reading);
+            if (number == null) {
+                activity.showErrorMessage();
             } else {
-                activity.finishActivity();
+                boolean isReadingAdded;
+                if ("mg/dL".equals(getUnitMeasuerement())) {
```
emartynov: That is the third nested level in the logic flow. How to about extract some logic to methods (like `createReading`) to make it more readable?

app/src/test/java/org/glucosio/android/tools/ReadingToolsTest.java
```diff
@@ -0,0 +1,32 @@
...
+    }
+
+    @Test
+    public void shouldReturnNull_whenInvalidStringPassed() {
+        assertThat(ReadingTools.parseReading("abc")).isNull();
+    }
+
+    @Test
+    public void shouldReturnCorrectNumber_whenDotIsDecimalSeparator() {
+        Locale.setDefault(new Locale("en"));
```
emartynov: That would influence all following tests in the suite. Should we return original local in `@After`?

app/src/test/java/org/glucosio/android/tools/ReadingToolsTest.java
```diff
@@ -0,0 +1,32 @@
+package org.glucosio.android.tools;
+
+import org.glucosio.android.RobolectricTest;
+import org.junit.Test;
+
+import java.util.Locale;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class ReadingToolsTest extends RobolectricTest {
```
emartynov: Should it be really `RobolectricTest`? Isn't plain junit enough?

app/src/test/java/org/glucosio/android/tools/ReadingToolsTest.java
```diff
@@ -0,0 +1,32 @@
...
+
+    @Test
+    public void shouldReturnCorrectNumber_whenDotIsDecimalSeparator() {
+        Locale.setDefault(new Locale("en"));
+        assertThat(ReadingTools.parseReading("1.23").doubleValue()).isEqualTo(1.23);
+    }
+
+    @Test
+    public void shouldReturnCorrectNumber_whenCommaIsDecimalSeparator() {
+        Locale.setDefault(new Locale("fr"));
```
emartynov: Same comment about locale

prt2121:
Updated my PR: using junit, reseting locale after tests, and cleaning up `AddReadingPresenter`.
Thanks for your thorough review.

app/src/main/res/layout/activity_add_glucose.xml
prt2121: this allows `,` to show on soft keyboard. as far as i test, it's required. more details at
http://stackoverflow.com/questions/3821539/decimal-separator-comma-with-numberdecimal-inputtype-in-edittext

emartynov: Clear! Thank you!

emartynov:
@prt2121 thank you a lot for the contribution!
LGTM 👏🎩

prt2121:
👍

PR #313, opened 2016-10-13
Add mmol/L unit check in validGlucose method

racell:
mmol/L is already in use. But in AddGlucosePresenter.java file, mmol/L was not approved for its validation. 

It should be tested for its validation and I worked on it.


bkerensa:
@emartynov Can you review this?

raacker:
LGTM

Issue #303, opened 2016-09-25
Make sure all strings like "mmol/L" and "mg/dL" use localized versions globally


TonnesM:
Glucosio currently uses several instances of strings like "mmol/L" or "mg/dL" that may require lowercase "l" for some locales. This was reported earlier in e.g. https://crowdin.com/translate/glucosio/all/en-enus#6268 but only partly fixed, i.e. in Settings, not the overview or elsewhere.

Please make sure such instances get fixed in the entire app.

emartynov:
Should be closed?

PR #314, opened 2016-10-13
Remove hardcoded strings for concentrations like mg/dL

apicellaj:
This addresses the issue of hardcoded strings mentioned in issue #303

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -443,6 +443,10 @@
     <string name="activity_backup_drive_button_open_drive">Manage on Drive</string>
     <string name="mmol_mol">mmol/mol</string>
     <string name="mmol_L">mmol/L</string>
+    <string name="mmol_mol_value">%1$s mmol/mol</string>
+    <string name="mmol_L_value">%1$s mmol/L</string>
+    <string name="mg_dL">mg/dL</string>
+    <string name="mg_dL_value">%1$s mg/dL</string>
     <string name="preferences_graph_old">Revert to old graph</string>
     <string name="history">history</string>
     <string name="save">save</string>
--- a/app/src/main/java/org/glucosio/android/activity/AddA1CActivity.java
+++ b/app/src/main/java/org/glucosio/android/activity/AddA1CActivity.java
@@ -62,7 +62,7 @@ public class AddA1CActivity extends AddReadingActivity {
         this.createFANViewAndListener();
 
         if (!"percentage".equals(presenter.getA1CUnitMeasuerement())) {
-            unitTextView.setText("mmol/mol");
+            unitTextView.setText(getString(R.string.mmol_mol));
         }
 
         // If an id is passed, open the activity in edit mode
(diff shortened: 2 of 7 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -443,6 +443,10 @@
     <string name="activity_backup_drive_button_open_drive">Manage on Drive</string>
     <string name="mmol_mol">mmol/mol</string>
     <string name="mmol_L">mmol/L</string>
+    <string name="mmol_mol_value">%1$s mmol/mol</string>
```
emartynov: Should it be some number instead of string? I mean `d` or `f` instead of `s`

app/src/main/res/values/strings.xml
```diff
@@ -443,6 +443,10 @@
...
+    <string name="mmol_L_value">%1$s mmol/L</string>
```
emartynov: Same question as above

app/src/main/res/values/strings.xml
```diff
@@ -443,6 +443,10 @@
...
+    <string name="mg_dL">mg/dL</string>
+    <string name="mg_dL_value">%1$s mg/dL</string>
```
emartynov: Same question as above

app/src/main/res/values/strings.xml
emartynov: Ah no, looks like we operate with strings everywhere. But these number strings are they localised?

apicellaj:
Thanks for the feedback! I can write tests for these changes but, to test localization, it might be helpful to know which locales needed this change so I can check for the proper behavior. Unless you have a different testing method in mind.

emartynov:
Actually, it is a good point. I was thinking just to test that texts in TextViews are correct:
```
assertThat(textView1).hasText(activity.getStirng(R.string.mg_dL));
assertThat(textView2).hasText(activity.getStirng(R.string. mg_dL_value, String.valueOf(10.0)));
```
But this will be not stable test until we have localisation for these texts. I mean our build machine definitely has EN local and will pass these tests. However, for people who have Italian as the current locale, for example, these tests will fail.

app/src/main/res/values/strings.xml
apicellaj: Do you mean the `mmol_mol_value`, `mmol_L_value` and `mg_dL_value` number strings? I added them to support localization because I thought concatenating `glucoseReading + getString(R.string.mmol_L)` would be bad practice in a left-to-right language locale.

emartynov:
BTW, all decimals will be formatted not according to locale:

- http://stackoverflow.com/questions/2627739/why-java-double-print-format-differs-depending-on-using-string-format-or-not
- http://docs.oracle.com/javase/7/docs/api/java/lang/Double.html#toString(double)

app/src/main/res/values/strings.xml
emartynov: I mean http://docs.oracle.com/javase/7/docs/api/java/lang/Double.html#toString(double)

apicellaj:
Oh, good point. So would a change to `String.format(Locale.getDefault(), String, Object...)` be the way to handle decimal formatting?

emartynov:
Looks good, but no tests. @PaoloRotolo, up to you to merge it.

PaoloRotolo:
I *have to* find the time to learn how to write tests. :/

apicellaj:
@emartynov I tried including tests but I'm not sure how to get around the dependencies (Realm, presenters, etc.) in the adapters and fragments that I want to test. Do you have any suggestions for testing?

PR #315, opened 2016-10-14
Fix mmol and a1c bugs

PaoloRotolo:
A1C:
"Not sure but there might be a bug in Android HbA1c calculator - if set glucose units to mmol/L and HbA1c units to mmol/mol, save button saves as if units for HbA1c is percentage"

MMOL:
"When using mmol/L and edit a glucose value, the value is presented with the mg/DL, when saved without changing the value (you only change the time), then you have the mg/DL value as if it was mmol/L. 
Need to convert the stored value back to mmol/L before displaying it on the edit page."

### DIFF
--- a/app/src/main/java/org/glucosio/android/activity/AddGlucoseActivity.java
+++ b/app/src/main/java/org/glucosio/android/activity/AddGlucoseActivity.java
@@ -119,7 +119,11 @@ public class AddGlucoseActivity extends AddReadingActivity {
         if (this.isEditing()) {
             setTitle(R.string.title_activity_add_glucose_edit);
             GlucoseReading readingToEdit = presenter.getGlucoseReadingById(this.getEditId());
-            readingTextView.setText(readingToEdit.getReading() + "");
+            if (presenter.getUnitMeasuerement().equals("mg/dL")) {
+                readingTextView.setText(String.valueOf(readingToEdit.getReading()));
+            } else {
+                readingTextView.setText(String.valueOf(presenter.convertToMmol(readingToEdit.getReading())));
+            }
             notesEditText.setText(readingToEdit.getNotes());
             cal.setTime(readingToEdit.getCreated());
             this.getAddDateTextView().setText(dateTime.getDate(cal));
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/glucosio/android/activity/AddGlucoseActivity.java
```diff
@@ -119,7 +119,11 @@ public void onNothingChosen(View labelledSpinner, AdapterView<?> adapterView) {
         if (this.isEditing()) {
             setTitle(R.string.title_activity_add_glucose_edit);
             GlucoseReading readingToEdit = presenter.getGlucoseReadingById(this.getEditId());
-            readingTextView.setText(readingToEdit.getReading() + "");
+            if (presenter.getUnitMeasuerement().equals("mg/dL")) {
+                readingTextView.setText(String.valueOf(readingToEdit.getReading()));
+            } else {
+                readingTextView.setText(String.valueOf(presenter.convertToMmol(readingToEdit.getReading())));
```
emartynov: Small duplication, you should determine text to be displayed in `if\else` and just set text after.

app/src/main/java/org/glucosio/android/presenter/A1CCalculatorPresenter.java
```diff
@@ -72,7 +72,14 @@ public void checkGlucoseUnit() {
     }
 
     public void saveA1C(double a1c) {
-        HB1ACReading a1cReading = new HB1ACReading(a1c, new Date());
+        User user = dbHandler.getUser(1);
+        double finalA1c = a1c;
+        if (!"percentage".equals(user.getPreferred_unit_a1c())) {
+            GlucosioConverter converter = new GlucosioConverter();
+            finalA1c = converter.a1cIfccToNgsp(a1c);
```
emartynov: Do we have a dictionary somewhere? I understand a1c but what are CC and NGPS I don't know.

PaoloRotolo: IFCC is A1C in mmol/mol and NGSP is the same in percentage.

PaoloRotolo: BTW this is explained in Converter's comments.
`    public double a1cNgspToIfcc(double ngsp) {
        // percentage to mmol/mol
        // [NGSP - 2.152] / 0.09148
        return round((ngsp - 2.152) / 0.09148, 2);
    }

    public double a1cIfccToNgsp(double ifcc) {
        // mmol/mol to percentage
        // [0.09148 * IFCC] + 2.152
        return round((0.09148 * ifcc) + 2.152, 2);
    }`

app/src/main/java/org/glucosio/android/activity/AddGlucoseActivity.java
PaoloRotolo: Fixed.

PR #316, opened 2016-10-14
[RFC] Support 'Almost' translated languages

### LINKED ISSUE #297: Hello View and Settings display only English
How Can I add more languages to display in the languages select?

raacker:
This PR support more languages. So it fixes #297 

This application is used in various countries. And in Crowdin, some languages are already 100% translated or almost translated. But Glucosio is still support only English. These features are enhancement of it. And also, you can find Crowdin's API for translation status. But there are too many non-translated languages and it was useless to get it's all data. The API needs Auth key so until it is cleared, just maintaining available strings manually would be better. I already made code, but I will just keep it.

- Add available_languages.xml of 'almost' translated languages
- Add updateLocale code on MainPresenter when there is already user profile in DB
- Swap two codes order of calling MainPresenter for update MainActivities in start
- Add localeHelper variable for MainActivity
- Changed getLocalesWithTranslation to available languages
- Fixed bug in build.gradle. It should contain one more backslash

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -171,7 +171,7 @@ def getAllLocales() {
 
     fileTree("src/main/res").visit { FileVisitDetails details ->
         if (details.file.path.endsWith("strings.xml")) {
-            def languageCode = details.file.parent.tokenize('/').last().replaceAll('values-', '').replaceAll('-r', '-')
+            def languageCode = details.file.parent.tokenize('/\\').last().replaceAll('values-', '').replaceAll('-r', '-')
             languageCode = (languageCode == "values") ? "en" : languageCode;
             foundLocales.append("\"").append(languageCode).append("\"").append(",")
         }
--- a/app/src/main/java/org/glucosio/android/presenter/MainPresenter.java
+++ b/app/src/main/java/org/glucosio/android/presenter/MainPresenter.java
@@ -31,10 +31,14 @@ public class MainPresenter {
 
     public MainPresenter(MainActivity mainActivity, DatabaseHandler databaseHandler) {
         dB = databaseHandler;
-        Log.i("msg::", "initiated db object");
+        Log.i("msg::", "initiated dB object");
         if (dB.getUser(1) == null) {
             // if user doesn't exists start hello activity
             mainActivity.startHelloActivity();
+        } else {
+            // If user already exists, update user's preferred language and recreate MainActivity
+            mainActivity.getLocaleHelper().updateLanguage(mainActivity,
+                                                          dB.getUser(1).getPreferred_language());
         }
     }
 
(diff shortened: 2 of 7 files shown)


app/build.gradle
```diff
@@ -53,7 +53,7 @@ android {
             testCoverageEnabled true
 
             buildConfigField 'String', 'GOOGLE_ANALYTICS_TRACKER', '"UA-68882401-2"'
-            buildConfigField "String[]", "TRANSLATION_ARRAY", "new String[]{}"
+            buildConfigField "String[]", "TRANSLATION_ARRAY", getAllLocales()
```
emartynov: Please don't do this. We disabled it for debug build to speed up development process. If you need please keep this change local.

app/src/main/java/org/glucosio/android/presenter/MainPresenter.java
```diff
@@ -27,18 +27,22 @@
...
+        db = databaseHandler;
         Log.i("msg::", "initiated db object");
-        if (dB.getUser(1) == null) {
+        if (db.getUser(1) == null) {
             // if user doesn't exists start hello activity
             mainActivity.startHelloActivity();
+        } else {
+            // If user already exists, update user's preferred language and recreate MainActivity
+            mainActivity.getLocaleHelper().updateLanguage(mainActivity,
+                                                          db.getUser(1).getPreferred_language());
```
emartynov: We have such code in `Application.onCreate()`. Why do we need it here?

app/src/main/java/org/glucosio/android/tools/LocaleHelper.java
```diff
@@ -54,35 +54,21 @@ public void updateLanguage(@NonNull final Context context, @NonNull final String
...
-        // Sorry :/
-        String englishString = "Automatic backup";
 
-        for (String localString : languages) {
-            // For each locale, check if we have translations
-            Configuration conf = resources.getConfiguration();
-            Locale savedLocale = conf.locale;
-            conf.locale = getLocale(localString);
-            resources.updateConfiguration(conf, null);
+        String[] translatedLanguages = resources.getStringArray(R.array.available_languages);
```
emartynov: We are trying to avoid it this. It means you have to update this file manually each time. The point of whole this method was to pickup one string and check if it is translated in language.

raacker: Yes I know the point. But It is not good for adjusting all languages. There must be exact key to choose which language is ready for service. 

As I wrote in my Commit and PR, there is API of Crowdin and I already made for it. But I need project key to use this api.

app/build.gradle
raacker: Okay. I'll keep it

raacker: force committed on it

app/src/main/java/org/glucosio/android/tools/LocaleHelper.java
raacker: you should aware that if there is 1% translated language and it translated only "Automate Backup", It will be added to supporting language. It should be removed.

emartynov: Good! I didn't realise there is support for crowdin API. Would be great to automate everything!

Issue #228, opened 2016-05-05
Allow user to lock app with pin or password


bkerensa:
In celebration of #PasswordDay I am creating this issue and as we think about tour and splash screen UI we should create a lock screen for the app and add functionality to allow user to set a pin or password to add an extra layer of security over their app install so if their device is stolen or used by someone Glucosio can be locked independently of the OS.

satya164:
Why not just use third-party apps to lock Glucosio?

bkerensa:
@satya164 because thats a third party app they would need to install
and its becoming more common for health apps or finance apps to have their own locking
feature.

DravitLochan:
you need splash screen too?

danielgomezrico:
Will be cool to support fingerprint too

vineyugave:
PR created  : https://github.com/Glucosio/glucosio-android/pull/319

PR #319, opened 2016-10-26
Added Pin Screen for Security

vineyugave:
This is for issue #228 
Will need an UI expert to work on the UI. Have a basic layout up there.


emartynov:
Thanks a lot for contribution! But I don't think this adds too much to the security for the application. Even more it is actually not completed feature.

emartynov:
I also don't understand why we see all these merge commits in this pull request.

app/build.gradle
```diff
@@ -42,6 +42,12 @@ android {
         testInstrumentationRunner 'android.support.test.runner.AndroidJUnitRunner'
     }
 
+    signingConfigs {
+        debug {
+            storeFile file('../extra/debug.keystore')
```
emartynov: Why debug key is part of this PR?

app/build.gradle
```diff
@@ -54,6 +60,8 @@ android {
 
             buildConfigField 'String', 'GOOGLE_ANALYTICS_TRACKER', '"UA-68882401-2"'
             buildConfigField "String[]", "TRANSLATION_ARRAY", "new String[]{}"
+
+            signingConfig signingConfigs.debug
```
emartynov: Why signing is part of this PR?

app/client_id.json
```diff
@@ -0,0 +1,13 @@
+{
```
emartynov: Why API config is part of this PR?

app/src/main/java/org/glucosio/android/activity/AppLockActivity.java
```diff
@@ -0,0 +1,54 @@
...
+
+import org.glucosio.android.R;
+import org.glucosio.android.tools.TimeElapsed;
+
+import java.util.Calendar;
+
+/**
+ * Created by Viney Ugave on 10/25/16.
+ */
+public class  AppLockActivity extends AppCompatActivity {
```
emartynov: Coul you call format code for this file?

app/src/main/java/org/glucosio/android/activity/AppLockActivity.java
```diff
@@ -0,0 +1,54 @@
...
+
+    private static final String TAG  = "AppLockActivity";
+    private static TimeElapsed timeElapsedInstance;
```
emartynov: Why static?

app/src/main/java/org/glucosio/android/activity/MainActivity.java
```diff
@@ -77,7 +77,7 @@
 import uk.co.chrisjenx.calligraphy.CalligraphyContextWrapper;
 
 
-public class MainActivity extends AppCompatActivity implements DatePickerDialog.OnDateSetListener, ExportView {
+public class MainActivity extends AppLockActivity implements DatePickerDialog.OnDateSetListener, ExportView {
```
emartynov: Why only `MainActivity`? What about other activities in the app?

app/src/main/java/org/glucosio/android/tools/TimeElapsed.java
```diff
@@ -0,0 +1,45 @@
...
+ */
+
+public class TimeElapsed {
+
+    private static TimeElapsed _instance = null;
+    private static long startTime;
+    private static long endTime;
+
+    protected TimeElapsed() {
+        // Exists only to defeat instantiation.
```
emartynov: If you want really prevent class creation you should make it final and constructor private

app/src/main/java/org/glucosio/android/tools/TimeElapsed.java
```diff
@@ -0,0 +1,45 @@
...
+        }
+        return _instance;
+    }
+
+    public static long getStartTime() {
+        return startTime;
+    }
+
+    public static void setStartTime() {
+        TimeElapsed.startTime = Calendar.getInstance().getTimeInMillis();
```
emartynov: Why not just `System.currentMilleSecs`?

app/src/main/java/org/glucosio/android/tools/TimeElapsed.java
```diff
@@ -0,0 +1,45 @@
...
+
+import java.util.Calendar;
+
+/**Singleton Class to intiate start time and calculate the time elapsed
+ * Created by Viney Ugave on 10/25/16.
+ */
+
+public class TimeElapsed {
+
+    private static TimeElapsed _instance = null;
```
emartynov: Why underscore in variable name?

app/src/main/java/org/glucosio/android/tools/TimeElapsed.java
```diff
@@ -0,0 +1,45 @@
...
+            _instance = new TimeElapsed();
+        }
+        return _instance;
+    }
+
+    public static long getStartTime() {
+        return startTime;
+    }
+
+    public static void setStartTime() {
```
emartynov: Probably better name for the method would be `resetStartTime`

app/src/main/res/layout/activity_app_lock.xml
```diff
@@ -0,0 +1,36 @@
+<?xml version="1.0" encoding="utf-8"?>
+<RelativeLayout xmlns:android="http://schemas.android.com/apk/res/android"
```
emartynov: Why not `LinerLayout`?

app/src/main/res/values/strings.xml
```diff
@@ -462,6 +462,7 @@
     <string name="reminder_item_title_generic">It is time to do an entry in Glucosio!</string>
     <string name="reminder_item_title_generic2">Time to log your glucose level!</string>
     <string name="warning_internet_connection_required">Sorry but you need internet for this feature</string>
+    <string name="PIN_Message">Please enter your pin below ...</string>
```
emartynov: This string will get lint warning about UNICODE symbol usage instead of multi dots

vineyugave:
@emartynov merge commits are because I did not fork earlier .. and then when I did .. I had to merge it to the fork... Please find my response inline : 
There is no way how user can set PIN  
_Where should this take place ? During first app launch. The direction was not clear so did not add it._ 
There is no way how user will unlock app with PIN
_I just have to add an action on hitting enter. But since pin is not did not add it, this should not take too long._
PIN si not involved to the securing user data
_Need direction on where to store the pin ? In the SQL database ? Please clarify. Have not added it coz was not sure_ .. thanks..

app/src/main/res/layout/activity_app_lock.xml
```diff
@@ -0,0 +1,36 @@
...
+        android:text="@string/PIN_Message"
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:layout_marginTop="46dp"
+        android:id="@+id/textViewPinMessage"
+        android:autoText="false"
+        android:textAlignment="center"
+        android:textSize="18sp"
+        android:layout_alignParentTop="true"
+        android:layout_alignParentLeft="true"
```
emartynov: No needs for horizontal alignment since view takes entire width of parent

app/src/main/res/layout/activity_app_lock.xml
```diff
@@ -0,0 +1,36 @@
...
+        android:layout_alignParentStart="true" />
+
+    <EditText
```
emartynov: How about making PIN field as password?

app/src/main/java/org/glucosio/android/activity/AppLockActivity.java
```diff
@@ -0,0 +1,54 @@
...
+        super.onStart();
+    }
+
+    @Override
+    protected void onResume() {
+        super.onResume();
+        //check if app has passed a time threshold
+        if(timeElapsedInstance.getStartTime() != 0){
+            timeElapsedInstance.setEndTime(Calendar.getInstance().getTimeInMillis());
+            long threshold = timeElapsedInstance.getEndTime()-timeElapsedInstance.getStartTime();
```
emartynov: Why do you need to set time and immediately to get it?

app/src/main/java/org/glucosio/android/activity/AppLockActivity.java
```diff
@@ -0,0 +1,54 @@
...
+                setContentView(R.layout.activity_app_lock);
+            }else{
+            }
+        }
+
+    }
+
+    @Override
+    protected void onStop() {
+        super.onStop();
```
emartynov: Why to override and don't do anything?

app/src/main/java/org/glucosio/android/activity/AppLockActivity.java
```diff
@@ -0,0 +1,54 @@
...
+    @Override
+    protected void onStart() {
+        super.onStart();
+    }
+
+    @Override
+    protected void onResume() {
+        super.onResume();
+        //check if app has passed a time threshold
+        if(timeElapsedInstance.getStartTime() != 0){
```
emartynov: This condition is always true

app/src/main/java/org/glucosio/android/activity/AppLockActivity.java
```diff
@@ -0,0 +1,54 @@
...
+            timeElapsedInstance.setEndTime(Calendar.getInstance().getTimeInMillis());
+            long threshold = timeElapsedInstance.getEndTime()-timeElapsedInstance.getStartTime();
+            Log.d(TAG,"Threshold : "+threshold);
+            //Current timeout threshold set to 30s
+            if(threshold>30000){
```
emartynov: This is wrong algorithm. Every resume of the activity that happens after 30 seconds after create will "lock" it. Basically, it will happen every dialog message showing or minimise/maximise app.

vineyugave: That is true.. but I could not think of a better approach .. can you suggest a better way/algorithm.. thanks

vineyugave:
@emartynov The PR was you to review the approach and get feedback .. so that I can update it based on the feedback.. the most critical part of handling the lock out was implemented .. the remaining parts are trivial and should not take much longer and can be extended to all the views.. thanks..

emartynov:
You merged three branches in your branch. You should only merge one that you started from.

I would suggest take your diff, branch dev again and just apply your diff on it. This will remove all merge request commits from PR.

emartynov:
You might ask clarification in existing issue that you referenced as fixed in this PR. Or you might ask them in Glucosio slack.

vineyugave:
@emartynov cool got it.. can fix the diffs.. can you please comment on any better approach to lock out the user.. Would like to know your view point..

app/src/main/java/org/glucosio/android/activity/AppLockActivity.java
emartynov: It might be 30 secs of the app inactivity. But I don't know what will be the best way to check it. The timer should be reset on any user action.

vineyugave: The idea is to reset the timer when the user confirms the pass code ... I looked at the behavior at some of the banking apps that I use and that is what I observed... 

If there is a better approach do let me know.. In the mean time I can make the updates to the remaining parts and update the PR or create a new one based on the merge conflict issue.. I ll update by next week .. thanks..

vineyugave:
New PR : https://github.com/Glucosio/glucosio-android/pull/325

emartynov:
Closing as not relevant

PR #324, opened 2016-10-29
Solving the CSV Export problem on Samsung S6

Ghazghkull:
The main problem seems to be related witth the static method exporting the CSV.

Creating the needed class every time into the new thread solves the problem


satya164:
Why does the error happen with a static method?

Ghazghkull:
@satya164 , I think you can find the response here: [](http://stackoverflow.com/questions/17343157/static-method-behavior-in-multi-threaded-environment-in-java)

paolorotolo:
So I talked with @bkerensa. Even if we fully don't understand why the error occurs, a fix is better than nothing. @Ghazghkull I see a conflict (sorry :P), can you please solve it? I will merge after.

Ghazghkull:
This night when I can access to my personal computer, I will solve the conflict :)

paolorotolo:
I solved some problems with dependencies. Now you should be able to compile without issues :)

Ghazghkull:
Thanks, last night I had a lot of errors and I din't know why....

paolorotolo:
Merged, thanks!

PR #326, opened 2016-11-07
Modified error code of a1c converting

raacker:
I was writing test code of A1CCalculaterPresenter and there was a logical error.

if user prefers percentage, it should be convertedto mmol/mol.

but it is calling NgspToIfcc when it is not percentage.


paolorotolo:
Hey, is this still valid? Why did you close the pull?

raacker:
Hmm.. I understood wrong! This method should return a1c value and if user preferred mmol, it has to convert from percentage to mmol! 

Nope. It is invalid :)  so I swapped codes for more readability

paolorotolo:
Okay, thanks anyway :)

raacker:
Thanks! :)

PR #327, opened 2016-11-07
Change GlucosioConverter to final class and changed its static methods

raacker:
GlucosioConverter is just for calculating between each units.
But in every codes, converter instance is always generating.
It is unacceptable to make more unnessesary instances.

### DIFF
--- a/app/src/main/java/org/glucosio/android/presenter/AddGlucosePresenter.java
+++ b/app/src/main/java/org/glucosio/android/presenter/AddGlucosePresenter.java
@@ -43,13 +43,11 @@ public class AddGlucosePresenter extends AddReadingPresenter {
     private DatabaseHandler dB;
     private AddGlucoseActivity activity;
     private ReadingTools rTools;
-    private GlucosioConverter converter;
 
     public AddGlucosePresenter(AddGlucoseActivity addGlucoseActivity) {
         this.activity = addGlucoseActivity;
         dB = new DatabaseHandler(addGlucoseActivity.getApplicationContext());
         rTools = new ReadingTools();
-        converter = new GlucosioConverter();
     }
 
     public void updateSpinnerTypeTime() {
@@ -100,7 +98,7 @@ public class AddGlucosePresenter extends AddReadingPresenter {
         if ("mg/dL".equals(getUnitMeasuerement())) {
             readingValue = number.intValue();
         } else {
-            readingValue = converter.glucoseToMgDl(number.doubleValue());
+            readingValue = GlucosioConverter.glucoseToMgDl(number.doubleValue());
         }
         GlucoseReading gReading = new GlucoseReading(readingValue, type, finalDateTime, notes);
         if (oldId == UNKNOWN_ID) {
@@ -109,10 +107,6 @@ public class AddGlucosePresenter extends AddReadingPresenter {
             isReadingAdded = dB.editGlucoseReading(oldId, gReading);
         }
         return isReadingAdded;
-    }
-
-    public double convertToMmol(int mgDl) {
-        return converter.glucoseToMmolL(mgDl);
     }
 
     public Integer retrieveSpinnerID(String measuredTypeText, List<String> measuredTypelist) {
--- a/app/src/main/java/org/glucosio/android/tools/GlucosioConverter.java
+++ b/app/src/main/java/org/glucosio/android/tools/GlucosioConverter.java
@@ -23,7 +23,9 @@ package org.glucosio.android.tools;
 import java.math.BigDecimal;
 import java.math.RoundingMode;
 
-public class GlucosioConverter {
+public final class GlucosioConverter {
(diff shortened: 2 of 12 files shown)


emartynov:
Thanks for the contribution!

Just one question - Why "It is unacceptable to make more unnecessary instances"?

raacker:
Sure. Because When Some class needs to use GlucosioConverter, It should always instantiate new object. But when it finishes its work, It is useless object anymore. But in every classes, this thing happens always. Keep a class in simple and single is really important :) Just keeping only one object in whole program!

emartynov:
Thanks for the explanation, I'm on your side but I'm just playing "bad complaining lady".

I agree that this code might not be optimal as it can be. But:
1) Static might make unit testing harder than proper dependency injection
2) Solution to have static methods to have more efficient code might be also solved by dependency injection 
3) I didn't find which class became more simple and which object has only one instance in the code

raacker:
Hmm okay! Thanks for your explanation. 

But I'm focusing on what GlucosioConverter should do. It doesn't have any field and its own. It is just a tool and support some calculation between units. And it might be not much bigger in Glucosio but increasing garbages also affect on performance.

And yes. There are not simplified class but this app has only one calculater, GlucosioConverter and it makes more simple to others.

Because it is not more that calculator.

There are no dependency to this

emartynov:
Clear! Do we have tests to check that changes didn't break anything?

I don't mind to merge these changes unless @paolorotolo has some objections. Let wait for his reaction.

raacker:
Sure :) Actually, it only changes it's invokation so it would not break something. And when this finished, gonna PR the test of converter

raacker:
Maybe using enum would more better to keep singleton

app/src/main/java/org/glucosio/android/presenter/AddGlucosePresenter.java
```diff
@@ -112,7 +110,7 @@ private boolean createReading(String type, String notes, long oldId, Date finalD
     }
 
     public double convertToMmol(int mgDl) {
```
emartynov: Maybe inline this method?

app/src/main/java/org/glucosio/android/tools/GlucosioConverter.java
```diff
@@ -23,7 +23,7 @@
 import java.math.BigDecimal;
 import java.math.RoundingMode;
 
-public class GlucosioConverter {
+public abstract class GlucosioConverter {
```
emartynov: Why abstract?

raacker: ah it was final when I created this PR, but if it doesn't need to instantiate, blocking and preventing it would be more better to everyone who would use this class :)

app/src/main/java/org/glucosio/android/presenter/AddGlucosePresenter.java
raacker: Hmm.. inline ? how?? because other usages are also working like this. They should be all changed to inline! 

So would you recommend your thought?

emartynov: Having one line method in presenter that has almost same name as in converter doesn't have much sense. So I would inline method in presenter.

app/src/main/java/org/glucosio/android/tools/GlucosioConverter.java
emartynov: Final with private constructor will work better. Right now you can still extend it and instantiate child class.

raacker: ah Okay! gonna work on it. Thanks @emartynov :)

app/src/main/java/org/glucosio/android/presenter/AddGlucosePresenter.java
raacker: aha right! :)

emartynov:
Again something wrong happened with merging. Why do we need merge commits in pull request?

raacker:
Because what I already pull requested was using GlucosioConverter method. I had to change it

emartynov:
Did you pull from master, develop or your own branch?

raacker:
Yes pulled from develop.

emartynov:
Can you instead of pull use rebase? 

So revert last commit and after "rebase on top" instead of "merge".

raacker:
Okay gonna work on it

raacker:
wow Actually I didn't know that how to handle with other commits. Thanks @emartynov !! 👍

emartynov:
You're welcome! One comment about inline method is still there.

raacker:
I misunderstood what you said :) Found it and cleared. Thank you very well!

emartynov:
Thanks! LGTM

PR #329, opened 2016-11-07
Refector createFANViewAndListener

Ksj7:
For reviewers.
This function is called multiple times in your source code.
This refactoring might be simplified to call it.

### DIFF
--- a/app/src/main/java/org/glucosio/android/activity/AddReadingActivity.java
+++ b/app/src/main/java/org/glucosio/android/activity/AddReadingActivity.java
@@ -134,6 +134,7 @@ public abstract class AddReadingActivity extends AppCompatActivity implements Ti
                 AnimationTools.startCircularReveal(doneFAB);
             }
         };
+        doneFAB.postDelayed(fabAnimationRunnable, 600);
     }
 
     protected abstract void dialogOnAddButtonPressed();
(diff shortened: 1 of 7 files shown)


emartynov:
Thanks for contribution and simplifications!

Just one question - what is this file about `새 텍스트 문서.txt`?

Ksj7:
Sorry! I made mistake :) remove unnecessary file!

Ksj7:
Seems handler instance is unnecessary. Changed it to doneFAB invocation.

app/src/main/java/org/glucosio/android/activity/AddReadingActivity.java
```diff
@@ -134,6 +134,7 @@ public void run() {
                 AnimationTools.startCircularReveal(doneFAB);
             }
         };
+        doneFAB.postDelayed(fabAnimationRunnable, 600);
```
emartynov: Please don't forget to remove callbacks when activity is destroyed. This will ensure no memory leaks (even for such short period 600ms).

Ksj7: There are already `doneFAB.removeCallbacks(fabAnimationRunnable);` exist onDestroy().
Should I do more things?

emartynov: Sorry for missing that point! In this case everything looks good. Merging!

PR #331, opened 2016-11-08
Add Testcase to HelloPresenter

MinSungKing:
- Exsisting testcode dosen't check when age is empty


emartynov:
Thanks for contribution! Merging it!

PR #333, opened 2016-11-08
Remove duplicated method and change name of original one

raacker:
getGlucoseReadingById is calling getGlucoseReading. And getGlucoseReading is
just using id for its search! 

They are duplicated and just getGlucoseReadingById needed.

### DIFF
--- a/app/src/main/java/org/glucosio/android/db/DatabaseHandler.java
+++ b/app/src/main/java/org/glucosio/android/db/DatabaseHandler.java
@@ -289,7 +289,7 @@ public class DatabaseHandler {
         return readingList;
     }
 
-    public GlucoseReading getGlucoseReading(long id) {
+    public GlucoseReading getGlucoseReadingById(long id) {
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/glucosio/android/db/DatabaseHandler.java
```diff
@@ -402,7 +397,7 @@ public GlucoseReading getGlucoseReadingById(long id) {
         gReadings = GlucoseReading.getGlucoseReadingsWithZeros(whereString);
         int i;
         for (i=0; i < gReadings.size(); i++){
-            readings.add(gReadings.get(i).getGlucoseReading());
+            readings.add(gReadings.get(i).getGlucoseReadingById());
```
emartynov: This is probably wrong cahnge

raacker: right! fixed it :)

emartynov:
Thanks!

PR #334, opened 2016-11-09
Adding PowerMock library

raacker:
I was writing test case of some classes, mockito can't test static methods.
So I'm trying to add PowerMock library.

I'm making PR to ask. Would adding it great? or any other suggestion?

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -149,6 +149,12 @@ dependencies {
     }
     testCompile 'io.reactivex:rxjava:1.1.8'
 
+    // PowerMock
+    testCompile "org.powermock:powermock-module-junit4:${powerMockVer}"
+    testCompile "org.powermock:powermock-module-junit4-rule:${powerMockVer}"
+    testCompile "org.powermock:powermock-classloading-xstream:${powerMockVer}"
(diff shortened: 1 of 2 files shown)


emartynov:
Can you give example what are you trying to test that requires PowerMock?

raacker:
Ah it is depend on if GlucosioConverter changed to utility class.

emartynov:
Just merged

raacker:
Okay then GlucosioConverter has static method and mockito can't test static methods.

http://stackoverflow.com/questions/21105403/mocking-static-methods-with-mockito

and that's what I'm gonna trying to code :)

emartynov:
Mockito is not testing anything!

You don't need to mock static methods just use the same method in the test.

raacker:
Oh ... hmm then can you suggest how to test this method? I don't know what I can do with this :)

[saveA1C method](https://github.com/Glucosio/glucosio-android/blob/develop/app/src/main/java/org/glucosio/android/presenter/A1CCalculatorPresenter.java#L73)

raacker:
oh... okay! got this :) I think powermock doens't need. Need to close this PR :)

emartynov:
I think so! Please close it!

PR #337, opened 2016-11-10
Add fully testcase to A1cCalculator

raacker:
I already tried to write all testcase but I had no idea how to test remaining methods. 

Finished it

### DIFF
--- a/app/src/test/java/org/glucosio/android/presenter/A1CCalculatorPresenterTest.java
+++ b/app/src/test/java/org/glucosio/android/presenter/A1CCalculatorPresenterTest.java
@@ -13,6 +13,7 @@ import org.mockito.runners.MockitoJUnitRunner;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.mockito.Matchers.anyLong;
 import static org.mockito.Mockito.when;
+import static org.mockito.Mockito.verify;
 
 @RunWith(MockitoJUnitRunner.class)
 public class A1CCalculatorPresenterTest {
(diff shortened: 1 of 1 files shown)


app/src/test/java/org/glucosio/android/presenter/A1CCalculatorPresenterTest.java
```diff
@@ -66,4 +72,24 @@ public void ShouldCalculatable_WhenUserPreferredPercentage() throws Exception {
...
+    }
+
+    @Test
+    public void ShouldReturnCorrectA1CUnit_WhenGetterCalled() throws Exception {
+        when(userMock.getPreferred_unit_a1c()).thenReturn("mmol/mol");
+        assertThat(presenter.getA1cUnit()).isEqualTo("mmol/mol");
+    }
+
+    @Test
+    public void ShouldSaveA1CCorrectly_WhenUserPreferredMmolForA1C() throws Exception {
```
emartynov: Better naming probably - should finish activity after saving

raacker: right! my mistake. thank you for pointing out

emartynov:
Could you also remove unused imports?

raacker:
eeh.. yeah so much issues from codacy :) Okay!

emartynov:
Thanks! Why do you rewrite history all the time? Why not just add new commits to pull request?

raacker:
Ah because just keep adding tons of commits with small changes make total history dirty! Maybe make new commit all the time and squash them all after approved, it would be best case. But just for case. Make commit simple and valuable is my commit style

emartynov:
There are no tons of commits. And we usually squash all changes in one commit per PR. 

You current style with one commit and history rewriting makes harder to me follow the changes.

emartynov:
Thanks! Merging!

raacker:
Okay then. I'll make all commit for each changes :)   thank you!

PR #338, opened 2016-11-10
Add fully testcase of GlucosioConverter

raacker:
Now, GlucosioConverter changed to static method and it doesn't have testcases. 

I tried to calculate values with other formulas. So It would be correct.

### DIFF
--- a/app/src/test/java/org/glucosio/android/tools/GlucosioConverterTest.java
+++ b/app/src/test/java/org/glucosio/android/tools/GlucosioConverterTest.java
@@ -1,0 +1,57 @@
+package org.glucosio.android.tools;
+
+import org.junit.Test;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class GlucosioConverterTest {
+    @Test
+    public void ShouldRoundToFloor_WhenLowerValueOfExactPointIsGiven(){
+        assertThat(GlucosioConverter.round(10.123, 2)).isEqualTo(10.12);
+    }
+
+    @Test
+    public void ShouldRoundToCeil_WhenHigherValueOfExactPointIsGiven(){
+        assertThat(GlucosioConverter.round(10.126, 2)).isEqualTo(10.13);
+    }
+
+    @Test
+    public void ShouldConvertMmolToMgDlExactly_WhenMgDlNeeded(){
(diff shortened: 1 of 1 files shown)


app/src/test/java/org/glucosio/android/tools/GlucosioConverterTest.java
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    @Test
+    public void ShouldConvertKgToLbExactly_WhenLbNeeded(){
+        assertThat(GlucosioConverter.kgToLb(60)).isEqualTo(132);
+    }
+
+    @Test
+    public void ShouldConvertlbToKgExactly_WhenKgNeeded(){
+        assertThat(GlucosioConverter.lbToKg(132)).isEqualTo(59);
```
emartynov: Why 132 pounds to kg is not 60?

app/src/test/java/org/glucosio/android/tools/GlucosioConverterTest.java
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    @Test
+    public void ShouldConvertMgDlToA1cExactly_WhenA1CNeeded(){
+        assertThat(GlucosioConverter.glucoseToA1C(18)).isEqualTo(2.25);
+    }
+
+    @Test
+    public void ShouldConvertA1CToGlucoseExactly_WhenGlucoseNeeded(){
+        assertThat(GlucosioConverter.a1cToGlucose(2.25)).isEqualTo(17.88);
```
emartynov: Why not 18? Are we doing some wrong rounding?

raacker:
There is a bit problem with casting. You can find methods in GlucosioConverter that just casting double to int and floating point just cut out. I think there must be some more specific way to make values. 

those two problems are belong to it

emartynov:
Clear! I think we should add rounding instead of just taking int value from double

raacker:
Right. But it would not solve these problem totally :)

raacker:
I concerned about it and... I faced that it would be really tough change to make return value to double..

emartynov:
Do we have conclusion over it? Rounding or not? @paolorotolo

bkerensa:
@paolorotolo ?

PR #339, opened 2016-11-14
Remove duplicated method

IamSSam:
I think that getHB1ACReadingById method call getHB1ACReading and getWeightReadingById method call getWeightReading.
They are duplicated. So I rename them.


emartynov:
Great! Thank you for the contribution!

PR #341, opened 2016-11-15
Refactor AddReadingPresenter

jeonsoh:
- setReadingTimeNow and updateReadingSplitDateTime have duplicated codes
- getReadingCal and getReadingTime have same problem

### DIFF
--- a/app/src/main/java/org/glucosio/android/presenter/AddReadingPresenter.java
+++ b/app/src/main/java/org/glucosio/android/presenter/AddReadingPresenter.java
@@ -17,19 +17,6 @@ public class AddReadingPresenter {
     private String readingHour;
     private String readingMinute;
 
-    public void setReadingTimeNow() {
-        DateFormat inputFormat = new SimpleDateFormat("yyyy-MM-dd HH:mm");
-        Date formatted = Calendar.getInstance().getTime();
-
-        SplitDateTime addSplitDateTime = new SplitDateTime(formatted, inputFormat);
-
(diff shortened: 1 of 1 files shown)


emartynov:

Are these methods ever used? I see you changed methods itself but I don't see you changed places where they were used.

raacker:
Seems removed duplicated codes. There are some codes that have totally same functionality.

Seems Okay :)

jeonsoh:
Yes. raacker is right. 
I tend to just remove duplicated codes. So, not effect on any other codes.

app/src/main/java/org/glucosio/android/presenter/AddReadingPresenter.java
```diff
@@ -40,10 +27,9 @@ public void updateReadingSplitDateTime(Date readingDate) {
         this.readingMonth = splitDateTime.getMonth();
     }
 
-    public Date getReadingTime() {
-        Calendar cal = Calendar.getInstance();
-        cal.set(Integer.parseInt(readingYear), Integer.parseInt(readingMonth) - 1, Integer.parseInt(readingDay), Integer.parseInt(readingHour), Integer.parseInt(readingMinute));
-        return cal.getTime();
+    public void setReadingTimeNow() {
+        Date formatted = Calendar.getInstance().getTime();
```
emartynov: Just `new Date()` is enough

jeonsoh:
Okay! I modified that part, changed to new Date(). 
Thank you very well!

PR #342, opened 2016-11-27
Rewrite SO copied code

emartynov:
Rewrote or replaced code that was copied from SO.

Please review and comment.


paolorotolo:
Looks good.

PR #344, opened 2016-12-13
Reduce ReadingTools.java hourToSpinnerType code to understand easily

HyungJiny:
I think that the case if hour is over 23 is don't need if you using '%' operation.
And 'return 8' repeated.
I think my code can reduce repeat code.


bkerensa:
@paolorotolo

emartynov:
Hey! You're right that that case will never run `23 % 23 == 0` and we will never have 24 as hour value. However, I think it might be a bug also, since I expect app should predict sometimes night mode. The question is "when".

bkerensa:
@paolorotolo any thoughts on this?

emartynov:
I'm really sorry, but I'm closing PR (due to lack of review and attention to answer to this problem)

PR #357, opened 2017-04-28
Make FreeStyle Reading Activity protrait only.

jurrabi:
That way, there is no issue loosing a reading because the user changes orientation during the gesture of reading the sensor (happens to me all the time).

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -150,7 +150,9 @@
                 android:name="android.support.PARENT_ACTIVITY"
                 android:value="org.glucosio.android.activity.MainActivity" />
         </activity>
-        <activity android:name=".activity.FreestyleLibreActivity">
+        <activity
+            android:name=".activity.FreestyleLibreActivity"
+            android:screenOrientation="portrait">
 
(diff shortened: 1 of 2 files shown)


emartynov:
Thank you a lot for the contribution!

I'm fine with this temporary fix. However, we should fix this issue in a different manner. Reading should not be interrupted or lost by orientation change.

jurrabi:
Thanks not needed. I thank to you and all contributors for this fine app.

I agree with you that it is not the best solution. But as you say, it is better than the current situation.

I do plan to work and contribute more things to this great project. I'm not any expert on Android development (I wish) but I'll try to contribute the best way I can.

Thanks again.

emartynov:
@paolorotolo any objections?

jurrabi:
Sorry. This question is probably caused by my poor understanding on how git and github pull requests work. I see the status of the pull request as 4-done... but I don't see any change in the develop branch of Glucosio.

Am I missing something?

Also, I did another change in my fork to add 2 strings used in some translation file but not in the default strings file that were generating 2 warnings in each build...

I see that changed is also reflected in this pull request. Shall I avoid pushing any more changes to my forked develop branch until the pull request is finished?

Issue #351, opened 2017-03-02
Scanning values from FreeStyle Libre is always in mg/dL


trizt:
When you have set preferred glucose unit to mmol/L you expect that the value scanned from the FreeStyle Libre would be presented in mmol/L, but no it's presented as mg/dL and when you select save it's entered as mg/dL in the Concentration filed which supposed to be mmol/L and you can't save as the value is out of range.

bkerensa:
@paolorotolo can we get a fix for this and push a release soon?

paolorotolo:
On it.

emartynov:
Should be closed?

PR #359, opened 2017-06-01
Measurement unit fixes. Closes #349 #351

### LINKED ISSUE #349: Settings/Range Min/Max value don't support mmol/L
The range Min value and Max value has to be entered in mg/dL even if your preferred glucose unit is mmol/L.

Version 1.3.2

paolorotolo:
Please review and merge.


paolorotolo:
Fixes #349 #351

Issue #178, opened 2015-12-04
Color blind friendlier icons


Aeyoun:
Both Overview and History depends on colors to show high and lows. This works well for many users, but for many these colors all look the same! Use symbols as well as colors to represent values.

Lows should use ▲ for highs (with an thick line below the rectangle for hypoglycemia), ▼ for lows, and ● for regular values. These symbols are commonly used in other diabetic software and are quite self-explanatory.

bkerensa:
I'd prefer we add a high contrast mode or add accessibility settings versus changing the global UX for the end user. We will evaluate this issue at our next meeting.

bkerensa:
@PaoloRotolo For this we can use some of the request by @Aeyoun for guidance but the pref is for us to have a toggle in settings under an accessibility menu so we do not change the entire global experience for non-blind users but can accommodate those who are color blind or having vision problems.

We will need UX feedback on this and thanks @Aeyoun for the issue.

PR #367, opened 2017-09-17
Glucose range display fixes

### LINKED ISSUE #91: Suggested blood glucose ranges
We should make a better description of blood glucose ranges suggested by different organizations. I know full titles are difficult to fit in UI design but I also believe abbreviation like ADA, AACE, UK Nice is not friendly to those who are not in US or UK. 

Also I'm not pretty sure if blood glucose ranges differs for people in different regions. For example TDA suggests to keep a 60-110mg/dl fasting glucose and < 140mg/dl glucose for 2 hours after meal, which looks quite different to what ADA and AACE suggests.

### LINKED ISSUE #289: Preferred range not displayed in Settings when selected
(originated from issue #288)

In Settings screen, the Preferred range does not display what's been selected, nor if that's remembered correctly. Any change in units may also fail since the Min and Max value don't change / remain grayed out when selecting Custom range.

### LINKED ISSUE #363: GlucoseRanges: Unaccessible 'blue' color value for 'ADA' preffered Range
Hi, 

I've already noticed that every value smaller than `70` would return `purple` so it won't return in any case`blue` color value for `ADA` preferred range, because conditional for this expects values lower than `70` which is as already reserved.

Here is an interesting part of code: 
```
 if (reading < 70) {
            return "purple";
        } else if (reading > 200) {
            return "red";
        } else if (reading > 70 | reading < 200) {
            // if not check with custom ranges
            switch (preferredRange) {
                case "ADA":
                    if (reading >= 70 & reading <= 180) {
                        return "green";
                    } else if (reading < 70) {
                        return "blue";
                    } else if (reading > 180) {
                        return "orange";
                    }
```

Check this code for `69`, `70` and `71` if that what was said is still not clear.

Cheers!

riskysnail:
* display both date and time for last reading in overview
* display the preferred range choice in preferences. Closes #289
* updates to GlucoseRanges. range presets are now displayed in user preferences as they are selected. Resolves #91
* change minimum glucose ranges for ADA and UK NICE presets. Closes #363

### DIFF
--- a/app/src/main/java/org/glucosio/android/tools/GlucoseRanges.java
+++ b/app/src/main/java/org/glucosio/android/tools/GlucoseRanges.java
@@ -34,17 +34,17 @@ public class GlucoseRanges {
     private String preferredRange;
     private int userMin;
     private int userMax;
-    private final static String ADA = "ADA";
-    private final static String AACE = "AACE";
-    private final static String UKNICE = "UK NICE";
-    private final static int HYPER_LIMIT = 200;
-    private final static int HYPO_LIMIT = 70;
-    private final static int ADA_MIN = 80;
-    private final static int ADA_MAX = 180;
-    private final static int AACE_MIN = 110;
-    private final static int AACE_MAX = 140;
-    private final static int UKNICE_MIN = 81;
-    private final static int UKNICE_MAX = 153;
+    private static final String ADA = "ADA";
+    private static final String AACE = "AACE";
+    private static final String UKNICE = "UK NICE";
+    private static final int HYPER_LIMIT = 200;
+    private static final int HYPO_LIMIT = 70;
+    private static final int ADA_MIN = 80;
+    private static final int ADA_MAX = 180;
+    private static final int AACE_MIN = 110;
+    private static final int AACE_MAX = 140;
+    private static final int UKNICE_MIN = 81;
+    private static final int UKNICE_MAX = 153;
 
     public GlucoseRanges(Context context) {
         this.mContext = context;
--- a/app/src/main/java/org/glucosio/android/tools/GlucoseRanges.java
+++ b/app/src/main/java/org/glucosio/android/tools/GlucoseRanges.java
@@ -63,13 +63,13 @@ public class GlucoseRanges {
     public static int rangePresetToMin(String preset) {
         switch (preset) {
             case "ADA":
-                return 70;
+                return 80;
             case "AACE":
                 return 110;
             case "UK NICE":
-                return 72;
+                return 81;
             default:
-                return 70;
+                return 80;
         }
     }
 
--- a/app/src/main/java/org/glucosio/android/tools/GlucoseRanges.java
+++ b/app/src/main/java/org/glucosio/android/tools/GlucoseRanges.java
@@ -32,17 +32,26 @@ public class GlucoseRanges {
     private DatabaseHandler dB;
     private Context mContext;
     private String preferredRange;
-    private int customMin;
-    private int customMax;
-    private final static int hyperLimit = 200;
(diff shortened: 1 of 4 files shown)


riskysnail:
Do I need to do anything about the coveralls check fail?

emartynov:
Sorry a lot for the delay with answer. I will try today to clean up warnings and finish the review.

app/src/main/java/org/glucosio/android/tools/GlucoseRanges.java
```diff
@@ -53,56 +61,57 @@ void setPreferredRange(String preferredRange) {
...
+    public static int getPresetMin(String preset) {
+        switch (preset) {
+            case ADA:
+                return ADA_MIN;
+            case AACE:
+                return AACE_MIN;
+            case UKNICE:
+                return UKNICE_MIN;
+            default:
+                return ADA_MIN;
```
emartynov, line 81: Why ADA is selected as default?

app/src/main/java/org/glucosio/android/tools/GlucoseRanges.java
```diff
@@ -53,56 +61,57 @@ void setPreferredRange(String preferredRange) {
...
 
     @VisibleForTesting
     void setCustomMax(int customMax) {
-        this.customMax = customMax;
+        this.userMax = customMax;
+    }
+
+    public static int getPresetMin(String preset) {
+        switch (preset) {
+            case ADA:
```
emartynov, line 74: Super minor: You can combine with `default` case

app/src/main/java/org/glucosio/android/tools/GlucoseRanges.java
```diff
@@ -53,56 +61,57 @@ void setPreferredRange(String preferredRange) {
...
+    public static int getPresetMax(String preset) {
+        switch (preset) {
+            case ADA:
+                return ADA_MAX;
+            case AACE:
+                return AACE_MAX;
+            case UKNICE:
+                return UKNICE_MAX;
+            default:
+                return ADA_MAX;
```
emartynov, line 94: Same question

emartynov (approved):
Looks good to me. Just one question about ADA is the default.

app/src/main/java/org/glucosio/android/tools/GlucoseRanges.java
riskysnail: It seemed reasonable. The default is only returned when the preset string given is not a known preset. It may be preferable to raise an error instead.

app/src/main/java/org/glucosio/android/tools/GlucoseRanges.java
riskysnail: Perhaps that is a better solution to the question below. Remove one of the cases and use default instead?

app/src/main/java/org/glucosio/android/tools/GlucoseRanges.java
emartynov: Oke, clear

emartynov:
Would be super nice if we write some tests for these changes.

Also, from the overall overview, the `GlucoseRanges` has two responsibilities now:

1. Keep user input
2. Constant class for ranges

It would be preferable to split them. Maybe in the future.

riskysnail:
I agree about GlucoseRanges. I'm thinking about creating a new class to handle the formatting of glucose readings (auto-converting to the preferred unit, potentially adding colorblind symbols for #178). This could be a good candidate to move the colorFromReading and stringToColor functions.

ghost:
@riskysnail what is status now? are you going to rearrange this code as you suggested in last commit?

PR #378, opened 2017-11-19
Historynotes

### LINKED ISSUE #368: Repeating notes in history view
Occasionally I add a note to a glucose reading. I noticed recently while scrolling through the history that some of my notes are displaying on other readings even though I've only entered that specific note once.

I tested this further from a dev build and by creating a reading with a note then adding older readings until I was able to see the note duplicate. It took 12 readings to see the duplicate (first and last reading have the note, 10 readings in between don't).

This spacing (10 readings) does not seem to be consistent. In my real data I saw only 9 readings between some duplicate notes.

riskysnail:
fix for repeating history notes. fixes #368

### DIFF
--- a/app/src/main/java/org/glucosio/android/activity/PreferencesActivity.java
+++ b/app/src/main/java/org/glucosio/android/activity/PreferencesActivity.java
@@ -190,7 +190,6 @@ public class PreferencesActivity extends AppCompatActivity {
                 @Override
                 public boolean onPreferenceChange(Preference preference, Object newValue) {
                     updatedUser.setCountry(newValue.toString());
-
                     updateDB();
                     return false;
                 }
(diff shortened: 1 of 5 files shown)


ghost:
@emartynov what is status of this PR?

emartynov:
I'm reading the original issue and I'm not sure about made assumptions or findings.

riskysnail:
Let me know what is unclear. Its pretty straight forward to test and 100% reproducible. I suspect the number of readings between repeating notes is determined by the number of views that can be displayed on screen.

Fore more info see:
https://developer.android.com/training/material/lists-cards.html

The diff of this pull request shows how the adapter was not clearing the notes of the view that is about to be recycled (notes were added to views if they were in the database but never cleared).
https://github.com/Glucosio/glucosio-android/pull/378/files

riskysnail:
Repro steps:
- go to History tab
- add a "TEST" note to the most recent reading
- scroll down until you see your TEST note on another reading

This works best if you don't have any other notes on any other readings or the adapter may update the TEST note with another one.

The space between the TEST note and seeing it again is determined by the number of visible views on your devices screen. For my phone that is 6 views (5 visible, 1 partially visible). So I need to scroll down 6 or 7 readings (partial views count so going 1 further is best) and the bottom reading shows my TEST note.

emartynov:

PR #380, opened 2017-11-27
Updating project to newer libraries versions. Adapt to AS 3.x

rcd27:


app/build.gradle
```diff
@@ -129,12 +128,11 @@ dependencies {
         transitive = true
     }
 
-    // https://mvnrepository.com/artifact/org.apache.commons/commons-math3
-    compile files('libs/commons-math3-3.6.1.jar')
+    implementation 'org.apache.commons:commons-math3:3.6.1'
```
emartynov: Is it big library? Why do we need it?

app/build.gradle
```diff
@@ -44,7 +43,7 @@ android {
 
     buildTypes {
         debug {
-            minifyEnabled false
+            minifyEnabled true
```
emartynov: It is needed for security but I prefer to have it in different PR

app/src/main/java/org/glucosio/android/GlucosioApplication.java
```diff
@@ -42,7 +42,7 @@
 
 import uk.co.chrisjenx.calligraphy.CalligraphyConfig;
 
-public class GlucosioApplication extends Application {
+public class GlucosioApplication extends MultiDexApplication {
```
emartynov: Why do we need it here? Did we already hit 65k limit?

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,6 @@
-#Mon Mar 20 12:26:00 CET 2017
+#Mon Nov 27 09:42:36 MSK 2017
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-3.5-all.zip
+distributionUrl=https\://services.gradle.org/distributions-snapshots/gradle-4.4-20171031235950+0000-all.zip
```
emartynov: Please no snapshots

emartynov (changes requested):
Please react to comments

app/src/main/java/org/glucosio/android/GlucosioApplication.java
rcd27: That was a merge mistake. No need in multidex with proguard actually.

app/build.gradle
rcd27: Seems like we don't.

app/build.gradle
rcd27: With `minifyEnabled false` we have ` Total methods in app-debug.apk: 57762 (88.14% used)`
And with `true`: `Total methods in app-debug.apk: 34668 (52.90% used)`

emartynov: I know, I know, but 58k is below 65k

emartynov: Usually, you can not just switch on proguard, it requires checking if app still working as it was before, if we need to adjust default settings, if we want to add more specific to our project, etc

emartynov:
And we have failing unit tests now!

app/build.gradle
rcd27: Ok, got it.

rcd27:
Was HelloActivityTest working before?

app/build.gradle
```diff
@@ -129,25 +130,22 @@ dependencies {
...
     testCompile "junit:junit:${junitVer}"
     testCompile "org.mockito:mockito-all:${mockitoVer}"
     testCompile "com.squareup.assertj:assertj-android:${androidAssertJVer}"
-    testCompile("org.robolectric:robolectric:${robolectricVer}") {
+    testImplementation("org.robolectric:robolectric:3.5.1") {
         exclude group: 'commons-logging', module: 'commons-logging'
         exclude group: 'org.apache.httpcomponents', module: 'httpclient'
     }
-    testCompile("org.robolectric:shadows-support-v4:${robolectricVer}") {
+    testImplementation("org.robolectric:shadows-support-v4:3.3.2") {
```
emartynov, line 144: If you started moving to new android gradle plugin then why not change it everywhere?

app/src/test/java/org/glucosio/android/fragment/AssistantFragmentTest.java
```diff
@@ -3,10 +3,13 @@
 import org.glucosio.android.RobolectricTest;
 import org.junit.Before;
 import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.robolectric.RobolectricTestRunner;
 import org.robolectric.shadows.support.v4.SupportFragmentTestUtil;
 
 import static org.assertj.core.api.Assertions.assertThat;
 
+@RunWith(RobolectricTestRunner.class)
```
emartynov: This is not needed! Check the `RobolectricTest` class

emartynov (approved):

emartynov:
There are still two more comments. Could you address them?

app/src/test/java/org/glucosio/android/fragment/AssistantFragmentTest.java
rcd27: Omg, missed that it extends RobolectricTest.class

emartynov:
Looks good! Please continue with the contribution!

I'm going to run test first to see if the app works and if so I will accept it!

ghost (commented):
Looks good to me. Tests passed :-)

emartynov:
Quite solid contribution! Thank you!

I'm merging!

PR #381, opened 2017-12-04
Restore existing espresso tests. 

rcd27:
There are some espresso tests which do not work properly right now. This PR is about restoring them to pass again.


rcd27:
Well, HelloActivityTest was restored, however I think layout is broken. This is supposed to pass without scrolling to CheckBox. When I open activity_hello.xml check box is visible on the same screen without any scrolling.

emartynov:
Could you pull upstream and rebase your branch on top of it?

rcd27:
Really do not know what to do with that "Demeter" violation...

emartynov:
Thank you for the contribution

Issue #383, opened 2017-12-09
CVS Export doesn't escape commas and linebreaks


AndyScherzinger:
For testing of the CVS export I added 3 readings and knowing that I'll do an CVS import afterwards I added line breaks and commas in the Notes fields of the readings which then leads to a broken CVS export since they aren't escaped to you end up with several lines with completely different content differing from the first line that defines all the included columns. 

```
Date,Time,Concentration,Preferred Glucose unit,Measured,Notes
09/12/2017,16:05,600,mg/dL,Night,Test
Tets
, test
Test, test, test

,
, test, 0,0,tet

09/12/2017,08:51,200,mg/dL,After breakfast,Test, 0
Trst,8
09/12/2017,08:46,400,mg/dL,After lunch,Test
```

PR #397, opened 2018-01-21
Escape characters in CSV

emartynov:
It is fix for #383.

Changes:
1. Refactored CSV export
2. Escaped '\n', '\r', '"', ',' when generating CSV
3. Extracted one more string constant
4. Added some missing tests

No changes here:
1. Using async task
2. Resume export after user allows write permission

Please review and comment.

### DIFF
--- a/app/src/main/java/org/glucosio/android/presenter/ExportPresenter.java
+++ b/app/src/main/java/org/glucosio/android/presenter/ExportPresenter.java
@@ -34,9 +34,9 @@ import org.glucosio.android.db.DatabaseHandler;
 import org.glucosio.android.db.GlucoseReading;
 import org.glucosio.android.tools.ReadingToCSV;
 import org.glucosio.android.view.ExportView;
+import org.joda.time.DateTime;
 
 import java.io.File;
-import java.util.Calendar;
 import java.util.List;
 
 public class ExportPresenter {
@@ -47,6 +47,7 @@ public class ExportPresenter {
             Manifest.permission.READ_EXTERNAL_STORAGE,
             Manifest.permission.WRITE_EXTERNAL_STORAGE
     };
+
     private int fromDay;
     private int fromMonth;
     private int fromYear;
@@ -63,7 +64,7 @@ public class ExportPresenter {
         if (activity instanceof ExportView) {
             mExportView = (ExportView) activity;
         } else {
-            throw new RuntimeException("ExportPresenter Activity must implement ExportView interface");
+            throw new IllegalStateException("ExportPresenter Activity must implement ExportView interface");
         }
         dB = databaseHandler;
     }
@@ -83,16 +84,10 @@ public class ExportPresenter {
                     if (isExportAll) {
                         readings = dB.getGlucoseReadings(realm);
                     } else {
-                        Calendar fromDate = Calendar.getInstance();
-                        fromDate.set(Calendar.YEAR, fromYear);
-                        fromDate.set(Calendar.MONTH, fromMonth);
-                        fromDate.set(Calendar.DAY_OF_MONTH, fromDay);
+                        DateTime fromDate = new DateTime(fromYear, fromMonth, fromDay, 0, 0);
+                        DateTime toDate = new DateTime(toYear, toMonth, toDay, 0, 0);
 
-                        Calendar toDate = Calendar.getInstance();
-                        toDate.set(Calendar.YEAR, toYear);
-                        toDate.set(Calendar.MONTH, toMonth);
-                        toDate.set(Calendar.DAY_OF_MONTH, toDay);
-                        readings = dB.getGlucoseReadings(realm, fromDate.getTime(), toDate.getTime());
+                        readings = dB.getGlucoseReadings(realm, fromDate.toDate(), toDate.toDate());
                     }
 
                     mExportView.onExportStarted(readings.size());
@@ -149,48 +144,24 @@ public class ExportPresenter {
         }
     }
 
-    public int getFromDay() {
-        return fromDay;
-    }
-
     public void setFromDay(int fromDay) {
         this.fromDay = fromDay;
-    }
-
-    public int getFromMonth() {
-        return fromMonth;
     }
 
     public void setFromMonth(int fromMonth) {
         this.fromMonth = fromMonth;
     }
 
-    public int getFromYear() {
-        return fromYear;
-    }
-
     public void setFromYear(int fromYear) {
         this.fromYear = fromYear;
-    }
-
-    public int getToDay() {
-        return toDay;
     }
 
     public void setToDay(int toDay) {
         this.toDay = toDay;
     }
 
-    public int getToMonth() {
-        return toMonth;
-    }
-
     public void setToMonth(int toMonth) {
         this.toMonth = toMonth;
-    }
-
-    public int getToYear() {
-        return toYear;
     }
 
     public void setToYear(int toYear) {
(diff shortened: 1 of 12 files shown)


emartynov:
@AndyScherzinger can you review?

app/src/main/java/org/glucosio/android/tools/ReadingToCSV.java
```diff
@@ -22,110 +22,105 @@
...
-            return null;
+        } else {
+            for (GlucoseReading reading : readings) {
+                writeLine(osw,
+                        this.dateTool.convertRawDate(reading.getCreated()),
+                        this.dateTool.convertRawTime(reading.getCreated()),
+                        formatter.format(GlucosioConverter.glucoseToMmolL(reading.getReading())),
+                        Constants.Units.MMOL_L,
+                        valueOrEmptyString(reading.getReading_type()),
+                        valueOrEmptyString((reading.getNotes())
```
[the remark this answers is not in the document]
emartynov: Fixed

app/src/main/java/org/glucosio/android/presenter/ExportPresenter.java
```diff
@@ -34,166 +35,144 @@
...
                     super.onPostExecute(filename);
-                    if (filename != null) {
-                        Uri uri = FileProvider.getUriForFile(mActivity.getApplicationContext(),
-                                mActivity.getApplicationContext().getPackageName() + ".provider.fileprovider", new File(filename));
-                        mExportView.onExportFinish(uri);
-                    } else if (isEmpty[0]) {
-                        mExportView.onNoItemsToExport();
+                    if (EMPTY.equals(filename)) {
+                        view.onNoItemsToExport();
+                    } else if (filename != null) {
```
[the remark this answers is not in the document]
emartynov: Ok, applied

app/src/main/java/org/glucosio/android/tools/ReadingToCSV.java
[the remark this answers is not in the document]
emartynov: Fixed

app/src/main/java/org/glucosio/android/presenter/ExportPresenter.java
```diff
@@ -34,166 +35,144 @@
...
-
-    public int getFromYear() {
-        return fromYear;
-    }
+            new ReadingToCSV(context, preferredUnit).createCSVFile(readings, osw);
 
-    public void setFromYear(int fromYear) {
-        this.fromYear = fromYear;
+            return file.getPath();
+        } catch (Exception e) {
```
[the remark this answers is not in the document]
emartynov: Fixed

AndyScherzinger:
@emartynov sure, I'll give the branch a test drive :)

app/src/main/java/org/glucosio/android/tools/ReadingToCSV.java
```diff
@@ -22,110 +22,104 @@
...
+    private void writeLine(@NonNull OutputStreamWriter osw, @NonNull String... values) throws IOException {
         for (int i = 0; i < values.length; i++) {
-            osw.append(values[i]);
+            osw.append(escapeCSV(values[i]));
             osw.append(i == values.length - 1 ? '\n' : ',');
         }
     }
+
+    private String escapeCSV(String value) {
+        if (containsSymbolsToEscape(value)) {
```
AndyScherzinger, line 109: ```SYMBOLS_TO_ESCAPE``` actually contains more symbols that get checked (and lead to a ```return true```) than the following line will actually escape.

AndyScherzinger:
Looks good to me code wise, just a minor comment. Will test the export later today after work.

AndyScherzinger:
Like mentioned in the in line comment only " and ' seem to be escaped, line breaks still exist and won't be escaped. This leads to new lines which will be treated as new rows, e.g. the following export:
```
Date,Time,Concentration,Preferred Glucose unit,Measured,Notes
1/22/18,21:37,25,mg/dL,After dinner,"Yet
Another
'
""
Test"
1/22/18,21:36,23.5,mg/dL,After dinner,"This, is
A
New line
Test"
```
When imported into MS Excel (komma delimiter, "-String key):
2018-01-22 21_53_03-mappe1 - excel

So I would guess an in-app import of it would still fail (don't know how to do an import :/)

emartynov:
We don't have import, only export :(

I was quickly checking https://tools.ietf.org/html/rfc4180. It looks like MS Excel is wrong with importing CSV.

I will dive more :)

AndyScherzinger:
I also send it to my gmail adress and opened it as preview within gmail and there it works fine... so yeah, maybe we just have to ignore Excel...

app/src/main/java/org/glucosio/android/tools/ReadingToCSV.java
emartynov: Can you elaborate more? I don't get the comment, sorry.

Did I make some logical error?

emartynov:
Conclusion on this PR?

app/src/main/java/org/glucosio/android/tools/ReadingToCSV.java
AndyScherzinger: I'll try :) containsSymbolsToEscape checks for a certain set of characters while not all of them get escaped then.

AndyScherzinger:
@emartynov fine by me and as discussed excel has its cvs issues but that is excel not glucose :)

app/src/main/java/org/glucosio/android/tools/ReadingToCSV.java
emartynov: Can you be more specific? What characters are not escaped?

AndyScherzinger: I'll try :)

```
 if (containsSymbolsToEscape(value)) {
    return "\"" + value.replace("\"", "\"\"") + "\"";
} 
```
escaping is done for ```"``` while the condition ```containsSymbolsToEscape(value)``` needs to be met of course. Thus I expected the condition to check for the same pattern. But it actually checks for:
```
private boolean containsSymbolsToEscape(String value) {
        for (char c : SYMBOLS_TO_ESCAPE) {
            if (value.indexOf(c) > -1) {
                return true;
            }
        }

        return false;
    }
```
while ```SYMBOLS_TO_ESCAPE``` contains:
```
private static final char[] SYMBOLS_TO_ESCAPE = new char[]{',', '"', '\n', '\r'};
```` 
So I would have expected the actual ascaping to also then escape the other "symbols to escape".

AndyScherzinger: other than that also when ```\n``` has been found the check return true and you are escaping for ```"```

AndyScherzinger:
Hi @emartynov I tried to explain it again, sorry for not getting it clearly explained :( Hope this time I got it right :)

app/src/main/java/org/glucosio/android/tools/ReadingToCSV.java
emartynov: If `,` or `\n` or `\r` is detected then escaping is wrapping the whole string in the single quotes `"`. However, if the string contains single quote it should be escaped as double quote `""`. I hope it helps.

AndyScherzinger: Ah 👍 
I didn't think of that, now makes send. Thanks for the explanation and your patience 😃

AndyScherzinger:
All fine by me. Any feedback @paolorotolo?

app/src/main/java/org/glucosio/android/tools/ReadingToCSV.java
emartynov: No issues, I also didn't know it by heart and should, of course, mention it in PR description.

paolorotolo:
I'll try it ASAP

paolorotolo:
Works fine in LibreOffice, good work

screenshot from 2018-01-29 15-35-52

paolorotolo (approved):

PR #400, opened 2018-02-04
Dependencies update

emartynov:
Changes:

1. I stole a nice way to keep dependencies clean from Jake's project https://github.com/JakeWharton/SdkSearch/blob/master/build.gradle.
2. Modified all build scripts to work with it correctly.
3. I've also cleaned up unused repositories, dependencies that are included by default.
4. Updated almost all dependencies to the latest.
5. Added one finally block to close resource if an error happens in BackupActivity and small cleanup in this file.
6. Also use latest gradle in wrapper now.

Please review and comment.


emartynov:
I'm checking coverage issue now.

ghost (commented):
Looks good to me. Just wondering about extracting `applicationId 'org.glucosio.android'` - as an example to `buildConfig.applicationId` to keep it all in one place.

AndyScherzinger:
I guess the application id isn't extracted since it is "fixed"

paolorotolo:
Looks good to me, awesome!

Issue #330, opened 2016-11-07
Using Espresso on Error message during editing preferences


raacker:
I was working on [HelloActivity](https://github.com/Glucosio/glucosio-android/blob/develop/app/src/main/java/org/glucosio/android/activity/HelloActivity.java#L185)

and I found that using espresso would be really great with this! 
http://blog.xebia.com/android-custom-matchers-in-espresso/

How about it?

emartynov:
Go for it! More quality tests is better!

ArunSelvam11:
i have given a pull request for this issue, please review it. 
https://github.com/Glucosio/glucosio-android/pull/402

PR #402, opened 2018-02-04
Issue #330 - Added the espresso test for invalid age error toast

ArunSelvam11:
Please review & merge the branch to develop.
Got this issue from codetriage.com.
Thank you.


emartynov:
Hey, thanks for the contribution. I'm checking it now!

paolorotolo:
I'll leave to @emartynov the final word on it, he's our tests guy :D

ghost (commented):
This merge request covers only one of possible cases:

``` 
   private boolean validateAge(String age) {
        if (TextUtils.isEmpty(age)) {
            return false;
        } else if (!TextUtils.isDigitsOnly(age)) {
            return false;
        } else {
            int finalAge = Integer.parseInt(age);
            return finalAge > 0 && finalAge < 120; //THIS ONE
        }
    }
```
The first case - no value - would be another case like `check_011__IfIEnterNoValueForAge`, so it does not matter for now. The second one and the last depend on what we mean by `wrongAge`.

For me, this test case is enough to merge, because it checks if there is any effect of wrong input - it should display `Toast` with message. This test checks it ;-)

emartynov:
Thank you for the contribution! Merging!

Issue #404, opened 2018-04-16
Add notice about Analytics data


paolorotolo:
From Google's documentation:

> You will give your end users proper notice about the implementations and features of Google Analytics you use (e.g. notice about what data you will collect via Google Analytics, and whether this data can be connected to other data you have about the end user). You will either get consent from your end users, or provide them with the opportunity to opt-out from the implementations and features you use.

paolorotolo:
Fix merged in ```develop```, thanks @chwalters!

PR #405, opened 2018-04-18
Possible fix for #404

chwalters:
Displays a dialog explaining how analytics are used and allows the user to opt-out.  Can also opt-out at any time from the preferences/settings screen.

### DIFF
--- a/app/src/main/java/org/glucosio/android/activity/HelloActivity.java
+++ b/app/src/main/java/org/glucosio/android/activity/HelloActivity.java
@@ -21,10 +21,14 @@
 package org.glucosio.android.activity;
 
 import android.content.Context;
+import android.content.DialogInterface;
 import android.content.Intent;
+import android.content.SharedPreferences;
 import android.graphics.drawable.Drawable;
 import android.os.Bundle;
+import android.preference.PreferenceManager;
 import android.support.v4.content.res.ResourcesCompat;
+import android.support.v7.app.AlertDialog;
 import android.support.v7.app.AppCompatActivity;
 import android.util.Log;
 import android.view.WindowManager;
@@ -100,10 +104,13 @@ public class HelloActivity extends AppCompatActivity implements HelloView {
 
         initStartButton();
 
-        Analytics analytics = application.getAnalytics();
-        analytics.reportScreen("Hello Activity");
+        showAnalyticsExplanationDialog();
+
         Log.i("HelloActivity", "Setting screen name: hello");
+
+
     }
+
 
     private void initLanguageSpinner(final LocaleHelper localeHelper) {
         localesWithTranslation = localeHelper.getLocalesWithTranslation(getResources());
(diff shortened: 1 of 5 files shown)


app/src/main/java/org/glucosio/android/analytics/GoogleAnalytics.java
```diff
@@ -28,15 +34,19 @@ public void init(@NonNull final Context context) {
 
     @Override
     public void reportScreen(@NonNull String name) {
-        mTracker.setScreenName(name);
-        mTracker.send(new HitBuilders.ScreenViewBuilder().build());
+        if (enabled) {
```
paolorotolo, line 37: Even if reportScreen is disabled, Analytics could send Google other statistics about the app.
We may consider using `setAppOptOut(true)` to disable Analytics at app-level.

app/src/main/java/org/glucosio/android/analytics/GoogleAnalytics.java
```diff
@@ -20,6 +23,9 @@ public void init(@NonNull final Context context) {
         mTracker = analytics.newTracker(BuildConfig.GOOGLE_ANALYTICS_TRACKER);
         mTracker.enableAdvertisingIdCollection(true);
 
+        SharedPreferences sharedPref = PreferenceManager.getDefaultSharedPreferences(context);
+        enabled = sharedPref.getBoolean("pref_analytics_opt_in", false);
+
         if (BuildConfig.DEBUG) {
             com.google.android.gms.analytics.GoogleAnalytics.getInstance(context).setAppOptOut(true);
```
paolorotolo, line 30: This is what we do for debug configuration. We can call setAppOptOut here if the opt out preference is setted.

app/src/main/res/values/strings.xml
```diff
@@ -281,6 +281,17 @@
     <string name="preferences_freestyle_libre_experimental">EXPERIMENTAL: Enable FreeStyle Libre® support</string>
     <string name="preferences_font_dyslexia_experimental">EXPERIMENTAL: Increase readability for users with dyslexia</string>
     <string name="preferences_font_dyslexia">Increase readability for users with dyslexia</string>
+    <string name="preferences_analytics_opt_in">Enable analytics</string>
+    <string name="allow">Allow</string>
+    <string name="optout">Opt-Out</string>
+    <string name="analytics_usage">Analytics usage</string>
+    <string name="analytics_usage_overview">
```
paolorotolo, line 288: @bkerensa could you please add a paragraph to address collecting data for debug puroposes with Crashlytics?

app/src/main/java/org/glucosio/android/activity/HelloActivity.java
```diff
@@ -100,11 +104,14 @@ protected void onCreate(Bundle savedInstanceState) {
 
         initStartButton();
 
-        Analytics analytics = application.getAnalytics();
-        analytics.reportScreen("Hello Activity");
+        showAnalyticsExplanationDialog();
```
paolorotolo, line 107: We're showing the notice in HelloActivity. This will affect new user only and not current users (since they will jump directly to MainActivity). @elioqoshi how can we handle this from UX prospective? Can we prompt analytics notice and opt-out to existing users right when they open Glucosio after the update?

paolorotolo (changes requested):

chwalters:
I will address these issues and create a new PR

Issue #407, opened 2018-04-23
Weight entry broken after unit change


chwalters:
I installed the app, added a weight.  Realized I wanted to switch units from kg to pounds.  So I deleted the first body weight reading.  Then, I went back to settings and changed units to pounds.  Then went back and added another reading and the value shows as some weird number after I entered it.  Probably the kg value but with pounds label and lots of decimal places.  Anyway, weight entry seems a little broken when changing units

To reproduce - install app, accept defaults for units

1. Add glucose reading - since weight is not available for first entry
2. Add weight entry, notice that default unit is kg
3. Delete entry
4. Go to settings, change preferred weight units to lbs
5. Go back and add a weight reading.
6. Save the weight reading
7. Go to history, body weight, click and hold the new weight reading
8. Editor shows wrong value but label indicates correct units

In my example, 175.5 lbs gets converted on the edit screen to 79.60555560595478

chwalters:
Believe fixed in 1.5

PR #408, opened 2018-04-24
fix-for-407

chwalters:
Proposed fixes for #407

### DIFF
--- a/app/src/main/java/org/glucosio/android/activity/AddA1CActivity.java
+++ b/app/src/main/java/org/glucosio/android/activity/AddA1CActivity.java
@@ -70,7 +70,7 @@ public class AddA1CActivity extends AddReadingActivity {
         if (this.isEditing()) {
             setTitle(R.string.title_activity_add_hb1ac_edit);
             HB1ACReading readingToEdit = presenter.getHB1ACReadingById(getEditId());
-            readingTextView.setText(readingToEdit.getReading() + "");
+            readingTextView.setText(new StringBuilder().append(this.numberFormat.format(readingToEdit.getReading())).append("").toString());
             Calendar cal = Calendar.getInstance();
             cal.setTime(readingToEdit.getCreated());
             this.getAddDateTextView().setText(formatDateTime.getDate(cal));
--- a/app/src/main/java/org/glucosio/android/activity/AddA1CActivity.java
+++ b/app/src/main/java/org/glucosio/android/activity/AddA1CActivity.java
@@ -24,7 +24,6 @@ import android.os.Bundle;
 import android.support.v7.widget.Toolbar;
 import android.widget.TextView;
 import android.widget.Toast;
-
 import org.glucosio.android.R;
 import org.glucosio.android.db.HB1ACReading;
 import org.glucosio.android.presenter.AddA1CPresenter;
@@ -41,7 +40,7 @@ public class AddA1CActivity extends AddReadingActivity {
     protected void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
         setContentView(R.layout.activity_add_hb1ac);
-        Toolbar toolbar = (Toolbar) findViewById(R.id.activity_main_toolbar);
+        Toolbar toolbar = findViewById(R.id.activity_main_toolbar);
 
         if (toolbar != null) {
             setSupportActionBar(toolbar);
@@ -55,8 +54,8 @@ public class AddA1CActivity extends AddReadingActivity {
         setPresenter(presenter);
         presenter.setReadingTimeNow();
 
-        readingTextView = (TextView) findViewById(R.id.hb1ac_add_value);
-        unitTextView = (TextView) findViewById(R.id.hb1ac_unit);
+        readingTextView = findViewById(R.id.hb1ac_add_value);
+        unitTextView = findViewById(R.id.hb1ac_unit);
 
         this.createDateTimeViewAndListener();
         this.createFANViewAndListener();
@@ -70,7 +69,7 @@ public class AddA1CActivity extends AddReadingActivity {
         if (this.isEditing()) {
             setTitle(R.string.title_activity_add_hb1ac_edit);
             HB1ACReading readingToEdit = presenter.getHB1ACReadingById(getEditId());
-            readingTextView.setText(new StringBuilder().append(this.numberFormat.format(readingToEdit.getReading())).append("").toString());
+            readingTextView.setText(this.numberFormat.format(readingToEdit.getReading()));
             Calendar cal = Calendar.getInstance();
             cal.setTime(readingToEdit.getCreated());
             this.getAddDateTextView().setText(formatDateTime.getDate(cal));
--- a/app/src/main/java/org/glucosio/android/activity/AddCholesterolActivity.java
+++ b/app/src/main/java/org/glucosio/android/activity/AddCholesterolActivity.java
@@ -24,7 +24,6 @@ import android.os.Bundle;
 import android.support.v7.widget.Toolbar;
 import android.widget.TextView;
 import android.widget.Toast;
-
 import org.glucosio.android.R;
 import org.glucosio.android.db.CholesterolReading;
 import org.glucosio.android.presenter.AddCholesterolPresenter;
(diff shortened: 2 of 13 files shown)


app/src/main/java/org/glucosio/android/activity/AddA1CActivity.java
```diff
@@ -70,7 +70,7 @@ protected void onCreate(Bundle savedInstanceState) {
         if (this.isEditing()) {
             setTitle(R.string.title_activity_add_hb1ac_edit);
             HB1ACReading readingToEdit = presenter.getHB1ACReadingById(getEditId());
-            readingTextView.setText(readingToEdit.getReading() + "");
+            readingTextView.setText(new StringBuilder().append(this.numberFormat.format(readingToEdit.getReading())).append("").toString());
```
emartynov: This pattern is not required. You don't need to concatenate an empty string. Since you don't concatenate anything then you don't need `StringBuilder` also. So simplification is
```
numberFormat.format(readingToEdit.getReading())
```

The original code is doing `a + ""` to force call `toString()` method that will just print value itself.

chwalters: I was following advice from Android Studio IDE, it suggested the pattern.  There seem to be a lot of warnings / advice from the IDE that we aren't following

emartynov: Yeah, it warns you about `a + ""` which is inefficient and also the bad pattern. But it doesn't warn you about values conversion :)

chwalters:
Pretty sure all comments have been addressed in this one

app/src/main/java/org/glucosio/android/activity/AddA1CActivity.java
emartynov: Chris, you can replace all code with builder just with `numberFormat` call

app/src/main/java/org/glucosio/android/activity/AddCholesterolActivity.java
```diff
@@ -70,9 +70,11 @@ protected void onCreate(Bundle savedInstanceState) {
...
             setTitle(R.string.title_activity_add_cholesterol_edit);
             CholesterolReading readingToEdit = presenter.getCholesterolReadingById(this.getEditId());
-            totalChoTextView.setText(readingToEdit.getTotalReading() + "");
-            LDLChoTextView.setText(readingToEdit.getLDLReading() + "");
-            HDLChoTextView.setText(readingToEdit.getHDLReading() + "");
+
+
+            totalChoTextView.setText(new StringBuilder().append(this.numberFormat.format(readingToEdit.getTotalReading())).append("").toString());
+            LDLChoTextView.setText(new StringBuilder().append(this.numberFormat.format(readingToEdit.getLDLReading())).append("").toString());
+            HDLChoTextView.setText(new StringBuilder().append(this.numberFormat.format(readingToEdit.getHDLReading())).append("").toString());
```
emartynov: Same here

app/src/main/java/org/glucosio/android/activity/AddKetoneActivity.java
```diff
@@ -64,7 +64,7 @@ protected void onCreate(Bundle savedInstanceState) {
         if (this.isEditing()) {
             setTitle(R.string.title_activity_add_ketone_edit);
             KetoneReading readingToEdit = presenter.getKetoneReadingById(this.getEditId());
-            readingTextView.setText(readingToEdit.getReading() + "");
+            readingTextView.setText(new StringBuilder().append(this.numberFormat.format(readingToEdit.getReading())).append("").toString());
```
emartynov: Same here

app/src/main/java/org/glucosio/android/activity/AddPressureActivity.java
```diff
@@ -69,8 +69,8 @@ protected void onCreate(Bundle savedInstanceState) {
             PressureReading readingToEdit = presenter.getPressureReadingById(this.getEditId());
 
             // set reading values
-            minPressureTextView.setText(readingToEdit.getMinReading() + "");
-            maxPressureTextView.setText(readingToEdit.getMaxReading() + "");
+            minPressureTextView.setText(new StringBuilder().append(this.numberFormat.format(readingToEdit.getMinReading())).append("").toString());
+            maxPressureTextView.setText(new StringBuilder().append(this.numberFormat.format(readingToEdit.getMaxReading())).append("").toString());
```
emartynov: Same here

app/src/main/java/org/glucosio/android/activity/AddWeightActivity.java
```diff
@@ -62,14 +66,18 @@ protected void onCreate(Bundle savedInstanceState) {
...
         // If an id is passed, open the activity in edit mode
         FormatDateTime formatDateTime = new FormatDateTime(getApplicationContext());
         if (this.isEditing()) {
             setTitle(R.string.title_activity_add_weight_edit);
             WeightReading readingToEdit = presenter.getWeightReadingById(this.getEditId());
-            readingTextView.setText(readingToEdit.getReading() + "");
+            double weightVal = readingToEdit.getReading();
+            if (needUnitConversion)
+                weightVal = GlucosioConverter.kgToLb(weightVal);
+            readingTextView.setText(new StringBuilder().append(this.numberFormat.format(weightVal)).append("").toString());
```
emartynov: Same here

app/src/main/java/org/glucosio/android/analytics/GoogleAnalytics.java
```diff
@@ -20,23 +23,36 @@ public void init(@NonNull final Context context) {
...
         }
+
+
     }
 
     @Override
     public void reportScreen(@NonNull String name) {
-        mTracker.setScreenName(name);
-        mTracker.send(new HitBuilders.ScreenViewBuilder().build());
+        if (enabled) {
```
emartynov, line 43: We already set the flag to Analytics itself. Do you need to have this check here?

app/src/main/java/org/glucosio/android/analytics/GoogleAnalytics.java
```diff
@@ -20,23 +23,36 @@ public void init(@NonNull final Context context) {
...
 
+        SharedPreferences sharedPref = PreferenceManager.getDefaultSharedPreferences(context);
+        enabled = sharedPref.getBoolean("pref_analytics_opt_in", false);
+
         if (BuildConfig.DEBUG) {
             com.google.android.gms.analytics.GoogleAnalytics.getInstance(context).setAppOptOut(true);
             Log.i("Glucosio", "DEBUG BUILD: ANALYTICS IS DISABLED");
+        } else {
+            if (!enabled) {
+                com.google.android.gms.analytics.GoogleAnalytics.getInstance(context).setAppOptOut(true);
```
emartynov, line 34: What is value by default? Should we always set it, just in case if the user enables it back

app/src/main/java/org/glucosio/android/activity/AddA1CActivity.java
emartynov: Done

app/src/main/java/org/glucosio/android/activity/AddCholesterolActivity.java
emartynov: Done

app/src/main/java/org/glucosio/android/activity/AddKetoneActivity.java
emartynov: Done

app/src/main/java/org/glucosio/android/activity/AddPressureActivity.java
emartynov: Done

app/src/main/java/org/glucosio/android/activity/AddWeightActivity.java
emartynov: Done

app/src/main/java/org/glucosio/android/analytics/GoogleAnalytics.java
emartynov: No, we don't need it

app/src/main/java/org/glucosio/android/analytics/GoogleAnalytics.java
emartynov: Reworked

emartynov:
LGTM!

Issue #409, opened 2018-04-25
Wrong units when graphing body weight readings


chwalters:
On the overview screen, select body weight.  If user settings are for imperial units (lbs) - this is ignored on the chart and values are shown in metric (kg).  To reproduce, add body weight readings in imperial units (lbs) and then try to chart them.  Desire behaviour is to show the chart using the settings chosen by the user.

PR #410, opened 2018-04-25
fix-for-409

chwalters:
Proposed fix for #409 - displays the chart of weight values in the user's preferred units

### DIFF
--- a/app/src/main/java/org/glucosio/android/activity/AddWeightActivity.java
+++ b/app/src/main/java/org/glucosio/android/activity/AddWeightActivity.java
@@ -76,7 +76,7 @@ public class AddWeightActivity extends AddReadingActivity {
             double weightVal = readingToEdit.getReading();
             if (needUnitConversion)
                 weightVal = GlucosioConverter.kgToLb(weightVal);
-            readingTextView.setText(this.numberFormat.format(weightVal));
+            readingTextView.setText(numberFormat.format(weightVal));
             Calendar cal = Calendar.getInstance();
             cal.setTime(readingToEdit.getCreated());
             this.getAddDateTextView().setText(formatDateTime.getDate(cal));
--- a/app/src/main/java/org/glucosio/android/fragment/OverviewFragment.java
+++ b/app/src/main/java/org/glucosio/android/fragment/OverviewFragment.java
@@ -532,10 +532,19 @@ public class OverviewFragment extends Fragment implements OverviewView {
     private LineData generateWeightData() {
         List<String> xVals = new ArrayList<>();
         List<Entry> yVals = new ArrayList<>();
+        boolean needUnitConversion = false;
+
+        if (!"kilograms".equals(presenter.getWeightUnitMeasuerement())) {
+            needUnitConversion = true;
+        }
 
         int k = 0;
         for (int i = presenter.getWeightReadings().size() - 1; i >= 0; i--) {
             float val = Float.parseFloat(presenter.getWeightReadings().get(i).toString());
+            if (needUnitConversion) {
+                val = (float)GlucosioConverter.kgToLb(val);
+            }
+
             yVals.add(new Entry(k, val));
             k++;
         }
--- a/app/src/test/java/org/glucosio/android/activity/HelloActivityTest.java
+++ b/app/src/test/java/org/glucosio/android/activity/HelloActivityTest.java
@@ -1,7 +1,8 @@
 package org.glucosio.android.activity;
 
+import android.content.DialogInterface;
 import android.content.res.Resources;
-
+import android.support.v7.app.AlertDialog;
 import org.assertj.core.util.Lists;
 import org.glucosio.android.RobolectricTest;
 import org.junit.Before;
@@ -9,6 +10,7 @@ import org.junit.Test;
 import org.mockito.ArgumentCaptor;
 import org.mockito.Captor;
 import org.robolectric.Robolectric;
+import org.robolectric.shadows.ShadowDialog;
 
 import java.util.Collections;
 import java.util.Locale;
@@ -30,7 +32,7 @@ public class HelloActivityTest extends RobolectricTest {
     private ArgumentCaptor<String> stringCaptor;
 
     @Before
-    public void setUp() throws Exception {
+    public void setUp() {
         initMocks(this);
 
         when(getLocaleHelper().getDeviceLocale()).thenReturn(new Locale("en"));
@@ -41,7 +43,7 @@ public class HelloActivityTest extends RobolectricTest {
     }
 
     @Test
-    public void ShouldBindView_WhenCreated() throws Exception {
+    public void ShouldBindView_WhenCreated() {
         assertThat(activity.languageSpinner).isNotNull();
         assertThat(activity.countrySpinner).isNotNull();
         assertThat(activity.ageTextView).isNotNull();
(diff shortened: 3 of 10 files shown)


emartynov:
Chris, why did you change only one place and not all similar everywhere?

chwalters:
Didn’t have time to fix all will do tomorrow sorry

app/src/main/java/org/glucosio/android/activity/AddWeightActivity.java
```diff
@@ -62,14 +66,18 @@ protected void onCreate(Bundle savedInstanceState) {
...
 
         // If an id is passed, open the activity in edit mode
         FormatDateTime formatDateTime = new FormatDateTime(getApplicationContext());
         if (this.isEditing()) {
             setTitle(R.string.title_activity_add_weight_edit);
             WeightReading readingToEdit = presenter.getWeightReadingById(this.getEditId());
-            readingTextView.setText(readingToEdit.getReading() + "");
+            double weightVal = readingToEdit.getReading();
+            if (needUnitConversion)
+                weightVal = GlucosioConverter.kgToLb(weightVal);
```
emartynov: Was it somehow forgotten?

app/src/main/java/org/glucosio/android/activity/HelloActivity.java
```diff
@@ -100,11 +104,14 @@ protected void onCreate(Bundle savedInstanceState) {
 
         initStartButton();
 
-        Analytics analytics = application.getAnalytics();
```
emartynov: You can move report analytics to `onStart` of activity and there it will use flag or not. After the dialog is closed it will call this method. This way you can preserve removed test.

app/src/main/java/org/glucosio/android/fragment/OverviewFragment.java
```diff
@@ -532,10 +532,19 @@ private LineData generateKetonesData() {
...
+
+        if (!"kilograms".equals(presenter.getWeightUnitMeasuerement())) {
+            needUnitConversion = true;
+        }
 
         int k = 0;
         for (int i = presenter.getWeightReadings().size() - 1; i >= 0; i--) {
             float val = Float.parseFloat(presenter.getWeightReadings().get(i).toString());
+            if (needUnitConversion) {
+                val = (float)GlucosioConverter.kgToLb(val);
```
emartynov, line 545: Was it forgotten?

emartynov: The conversion I mean

app/src/test/java/org/glucosio/android/activity/HelloActivityTest.java
```diff
@@ -40,11 +40,6 @@ public void setUp() throws Exception {
         activity = Robolectric.buildActivity(HelloActivity.class).create().get();
     }
 
-    @Test
-    public void ShouldReportAnalytics_WhenCreated() throws Exception {
```
emartynov: Removing test because you can not fix it is usually bad idea

emartynov (commented):
The changes overall look good. 

The main comment I would separate these changes into two or even three:
* opt out analytics and crashes
* formatting fixes
* conversion of the weight

The small comment that I would modify test and code to make it working. But it is up to you.

app/src/main/java/org/glucosio/android/activity/AddWeightActivity.java
chwalters: I don't understand this comment

app/src/main/java/org/glucosio/android/activity/HelloActivity.java
chwalters: I'm sorry I don't understand why this is needed

app/src/main/java/org/glucosio/android/fragment/OverviewFragment.java
chwalters: I don't understand this comment

app/src/test/java/org/glucosio/android/activity/HelloActivityTest.java
chwalters: I think the test is pretty pointless

app/src/main/java/org/glucosio/android/fragment/OverviewFragment.java
emartynov: Question was - we never had this functionality so we never had proper working weight for imperial countries, correct?

app/src/test/java/org/glucosio/android/activity/HelloActivityTest.java
emartynov: Please give pointness definition of the test

emartynov: I also said it is up to you. I do understand situation when you don't get any support from the platform devs and tests are not the simplest thing to write or modify

app/src/main/java/org/glucosio/android/activity/AddWeightActivity.java
emartynov: Sorry for that. The question was - we never had this functionality so we never had a proper working weight for imperial countries, correct?

app/src/main/java/org/glucosio/android/activity/HelloActivity.java
emartynov: We have a bunch of activity lifecycle events in Activity. We are interested in two `onCreate()` and `onStart()`.

`onCreate()` is called only once and there you show dialogue where the user decides to allow or not analytics and crashes.

`onStart()` is called a couple of times and one when the user closes the dialogue. 

So the proposal about moving analytics report to start method will allow to easily test it as well. However, this will bring side effect that this analytics might be reported a couple of times.

app/src/main/java/org/glucosio/android/activity/AddWeightActivity.java
chwalters: Correct, as far as I can tell it never worked for imperial countries

app/src/test/java/org/glucosio/android/activity/HelloActivityTest.java
chwalters: Testing whether the user chooses to report requires a UI test of some kind, I don't know how to do that here

app/src/main/java/org/glucosio/android/fragment/OverviewFragment.java
chwalters: Correct, as far as I can tell it never worked for imperial countries

app/src/main/java/org/glucosio/android/activity/HelloActivity.java
chwalters: Again, I don't understand the benefit or reason why this would be good to do, if it is really good perhaps you can change it to show me how it should work?

app/src/main/java/org/glucosio/android/activity/AddWeightActivity.java
emartynov: Clear! 👍

app/src/main/java/org/glucosio/android/fragment/OverviewFragment.java
emartynov: 👍

app/src/main/java/org/glucosio/android/activity/HelloActivity.java
emartynov: I will return test and modify it today

emartynov: Addressed

emartynov:
LGTM

emartynov:
Rebased branch so no conflicts with develop

PR #413, opened 2018-06-07
Summary: Remove e-mail support as an option.

chrispecoraro:
- Remove from support menu.
- Comment out related handler code.

### DIFF
--- a/app/src/main/java/org/glucosio/android/activity/MainActivity.java
+++ b/app/src/main/java/org/glucosio/android/activity/MainActivity.java
@@ -415,31 +415,34 @@ public class MainActivity extends AppCompatActivity implements DatePickerDialog.
         builder.setItems(getResources().getStringArray(R.array.menu_support_options), new DialogInterface.OnClickListener() {
             @Override
             public void onClick(DialogInterface dialog, int which) {
-                if (which == 0) {
-                    // Email
-                    Intent emailIntent = new Intent(Intent.ACTION_SENDTO, Uri.parse("mailto:hello@glucosio.org"));
-                    boolean activityExists = emailIntent.resolveActivityInfo(getPackageManager(), 0) != null;
 
-                    if (activityExists) {
-                        startActivity(emailIntent);
-                    } else {
-                        showSnackBar(getResources().getString(R.string.menu_support_error1), Snackbar.LENGTH_LONG);
-                    }
-                } else {
-                    // Forum
-                    String url = "http://community.glucosio.org/";
-                    Intent i = new Intent(Intent.ACTION_VIEW, Uri.parse(url));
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/glucosio/android/activity/MainActivity.java
```diff
@@ -415,31 +415,34 @@ public void openSupportDialog() {
...
-                    i.setPackage("com.android.chrome");
-                    try {
-                        startActivity(i);
-                    } catch (ActivityNotFoundException e) {
-                        // Chrome is probably not installed
-                        // Try with the default browser
-                        i.setPackage(null);
-                        startActivity(i);
-                    }
+//          E-mail support is no longer an option.
```
emartynov, line 419: Why keep commented code? If we need this code we can easily find it in the history

app/src/main/java/org/glucosio/android/activity/MainActivity.java
```diff
@@ -415,31 +415,34 @@ public void openSupportDialog() {
...
+//                        startActivity(emailIntent);
+//                    } else {
+//                        showSnackBar(getResources().getString(R.string.menu_support_error1), Snackbar.LENGTH_LONG);
+//                    }
+                // } else {
+                // Forum
+                String url = "http://community.glucosio.org/";
+                Intent i = new Intent(Intent.ACTION_VIEW, Uri.parse(url));
+                i.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK);
+                i.setPackage("com.android.chrome");
```
emartynov, line 436: Off topic, but why to ever set package here at all?

emartynov (changes requested):
I didn't test it but codewise it looks good. Could you remove commented code? And I would merge it as soon as testing confirms there is no regression or missed places where we do contact support more

app/src/main/java/org/glucosio/android/activity/MainActivity.java
chrispecoraro: I just thought that if we leave it, someday we want to re-enable e-mail as an option again, any new programmer could easily find this and just uncomment it.

app/src/main/java/org/glucosio/android/activity/MainActivity.java
chrispecoraro: I'm not familiar enough with Android to comment. @paolorotolo may know.

app/src/main/java/org/glucosio/android/activity/MainActivity.java
chwalters: In general keeping commented out code is not often a good idea, since the revision control system / history can show the change and what is commented out might not work again in the future so it might be confusing or cause problems down the line, just my 0.02

PR #414, opened 2018-06-07
Bumped build and version number (and version string)

chwalters:
Here I have bumped the version number, name and build number for your consideration


emartynov:
LGTM!

Issue #423, opened 2018-07-19
When using ',' as decimal separator readings are corrupted


chwalters:
When using phone locale such that ',' (comma) is the decimal separator (such as Nederland/Dutch), the app won't allow users to enter valid data, e.g., 12.8 becomes 128 when saved, but 12,8 isn't accepted as a valid input.  Reported by http://community.glucosio.org/t/cant-use-decimals-for-mmol-l-level/164/4 I was able to duplicate in Android simulator

Steps to Reproduce

Go to 'Settings->System->Languages & Input->Languages and change phone to 'Nederlands' locale.  Try adding a reading into Glucosio.  Readings cannot be edited correctly once the locale is changed either.

PR #424, opened 2018-07-19
develop-fix-423

chwalters:
Possible fixes for #423


emartynov:
LGTM! Did you test it?

chwalters:
I did run the 'tests' but also tried to test it on the simulator by switching different locales.  I tested a fair bit, but no doubt it could use even more

Issue #425, opened 2018-07-23
History list doesn't refresh when new data added


chwalters:
When viewing the history list for HbA1c, adding a new entry with the calculator works but the new reading isn't shown in the history list unless the user navigates away and comes back.  The new reading should show up in the history list after it is added.

PR #427, opened 2018-07-23
cnw-fix-425

chwalters:
Proposed fix for #425.  Reload history when main activity resumes

### DIFF
--- a/app/src/main/java/org/glucosio/android/activity/MainActivity.java
+++ b/app/src/main/java/org/glucosio/android/activity/MainActivity.java
@@ -283,14 +283,12 @@ public class MainActivity extends AppCompatActivity implements DatePickerDialog.
          */
         super.onResume();
 
-        if (viewPager != null) {
+        if (viewPager != null && homePagerAdapter != null) {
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/glucosio/android/activity/MainActivity.java
```diff
@@ -272,6 +274,28 @@ public boolean onItemClick(View view, int position, IDrawerItem drawerItem) {
...
+           example - while viewing HbA1c history, use calculator to add reading.  The new reading
+           wasn't being shown on the history until the user navigated away from the history page
+           and came back, and it should refresh with data IMHO
+         */
+        super.onResume();
+
+        if (viewPager != null && homePagerAdapter != null) {
+            int currentPage = viewPager.getCurrentItem();
+            if (currentPage == 1) {  // assumption - history tab is always position 1
+                HistoryFragment historyFragment = homePagerAdapter.getHistoryFragment();
```
emartynov, line 289: You can override onResume in the History fragment and reload data there. However, it is not a super great solution since it will happen also when the user locked the screen, the user opened another app, etc.

PR #432, opened 2021-12-28
data loss fix for adda1activity

Naitim:
Hello developers of glucosio

I am using your app glucosio. I think the app is great but I have one minor patch that could improve the user experience.

Here is a picture to help illustrate what activity that are changed in my patch:

https://ibb.co/f1bDCm0

When the user tries to create a new add HBa1c entry. If the screen focus goes to another app or activity(for example if an incoming call forces the user to go into the phone app), the user will lose any data they had put into this page if the app is force closed by Android.

This feature will automatically store the data when the user leaves the activity without submitting and restore said data when they restart the activity(this can be seen here where I closed the activity and restarted it https://ibb.co/RzN67dd). Therefore, the user does not have to fill in the data again thus improving the user experience.

I have tested out the feature to ensure it works.

If you have any questions or if you would like me to change anything, please do not hesitate to let me know!

Thank you for your time,
Tim

### DIFF
--- a/app/src/main/java/org/glucosio/android/activity/AddA1CActivity.java
+++ b/app/src/main/java/org/glucosio/android/activity/AddA1CActivity.java
@@ -20,6 +20,7 @@
 
 package org.glucosio.android.activity;
 
+import android.content.SharedPreferences;
 import android.os.Bundle;
 import android.support.v7.widget.Toolbar;
 import android.widget.TextView;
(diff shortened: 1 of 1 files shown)


emartynov:
Hi Tim,

Super thank you for the contribution! Unfortunately, The project was abandoned for quite some time already. I would recommend to fork the project or joining another active fork to avoid this PR staying open forever.

Eugen (The former contributor to the app)

emartynov:
BTW, for this case, it is better not to use shared preferences but check the `onSaveInstanceState` method.

